PR #22, opened 2015-01-09
support callable instance, instance method, and class method.

succhiello:
support below.

```python
class C(object):
    def __init__(self, arg):
        self._arg = arg
    def __call__(self, x):
        return x * self._arg
    def method(self, x):
        return x * self._arg * 10
    @classmethod
    def clsmethod(cls, x):
        return x * 100

rx.Observable.from_(range(10)).map(C(2)).subscribe(lambda x: print(x))
rx.Observable.from_(range(10)).map(C(2).method).subscribe(lambda x: print(x))
rx.Observable.from_(range(10)).map(C.clsmethod).subscribe(lambda x: print(x))
```

### DIFF
--- a/rx/internal/utils.py
+++ b/rx/internal/utils.py
@@ -1,4 +1,5 @@
 from inspect import getargspec
+import types
 
 from rx import AnonymousObservable
 from rx.disposables import CompositeDisposable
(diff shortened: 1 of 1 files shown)


dbrattli:
This is great work, thank you. I had been planning to do something with adapt_call, hopefully remove it since the `inspect` module is not available on micropython. One option is to simply use `try/except` and take the hit if `func`takes more than 1 param:

```python
def adapt_call(func):
    """Adapts func from taking n params to only taking 1 or 2 params"""
    
    def func1(arg1, *_):
        return func(arg1)
    def func2(arg1, arg2=None, *_):
        return func(arg1, arg2)

    def func_wrapped(*args, **kw):
        for fn in (func1, func2):
            try:
                return fn(*args, **kw)
            except TypeError:
                pass
            
    return func_wrapped
``` 

But I think I will just support both and check for `ImportError`for `inpect` and `types`. One additional fix we could make to your code is to accept constructors as well by checking for `__init__()`.  Here's some example code: http://pydoc.net/Python/rum/0.4dev-20110726/rum.util/ We could perhaps just use their implementation, but that code also takes some performance hit on every call. What do you think?

PR #23, opened 2015-01-18
BugFix. Pygame run error

zhuangzhuang:


dbrattli:
Would be nice with a patch that works for both 2.7 and 3.4. Maybe @enobayram could help out?

enobayram:
Sure, I'll look into it

enobayram:
Hi again, as far as I can see, zhuangzhuang's pull request is identical to the one I've submitted a few weeks ago. I've just tried to run the current "develop" branch on both 2.7 and 3.4 and it works as it is (which already contains zhuangzhuang's changes).

dbrattli:
Ok, what worried me was that this pull request reverted one part of your change:

```
-                 pos = event.__dict__["pos"]
+                pos = event.pos
```

But if you think everything is OK, then lets just leave it. Cannot test it myself. My PC stopped working, and I haven't been able to install pygame on my mac yet.

enobayram:
Ah, Ok. As I've said, zhuangzhuang's changes are identical to mine. As far as I can see, he has forked your master branch and made the fixes over it, that's why he didn't have those changes in the first place.

PR #24, opened 2015-02-13
Shortens and makes the konami code a bit easier to read.

jnlopar:
Just a tiny janitorial patch! :)


dbrattli:
This is nice! Thanks!

PR #27, opened 2015-03-01
Add examples/parallel/timer.py

38elements:
This is an example of using RxPY with `concurrent.futures`.


dbrattli:
Thanks a lot!

PR #29, opened 2015-03-12
Fix konamicode example

dano:
The current konamicode example is broken, because the `codes` list is missing two of the keys required to complete it properly. I just added the missing keys back in.


dbrattli:
Yes, looks like it was shortened a bit in the previous patch by @jnlopar

jnlopar:
Dang, good catch. I better go play more NES :(

PR #30, opened 2015-03-15
Fix comment in tofuture.py

38elements:
return => return_value
RSVP => Trollius

Trollius
http://trollius.readthedocs.org/en/latest/index.html


dbrattli:
Thanks!

PR #33, opened 2015-04-03
added Observable.with_latest_from()

pillmuncher:
Since I needed `Observable.with_latest_from()`, I built it. Maybe others might find it useful, too.


pillmuncher:
oops! forget to copy `tests/test_observable/test_withlatestfrom.py` into the right branch. Doing it now.

PR #34, opened 2015-04-03
added Observable.with_latest_from()

pillmuncher:
Since I needed `Observable.with_latest_from()`, I wrote it. Maybe others might find it useful, too.


dbrattli:
Nice work! Thanks a lot!

PR #37, opened 2015-04-10
Fix for #36: AsyncIOScheduler for Python 2

### LINKED ISSUE #36: AsyncIOScheduler for Python 2.7 via Trollius?
It seems possible to make the `AsyncIOScheduler` available for Python 2.7 through the use of [trollius](http://trollius.readthedocs.org/). Perhaps it can be configured through `rx.config`, like the class for `Future` can be?

hangtwenty:
Fix #36. Enable AsyncIOScheduler for Python 2 by using [trollius](http://trollius.readthedocs.org/) as `asyncio`. Also for a few test cases, use trollius's replacement for `yield from`, and [one of several possible replacements for `nonlocal`](http://stackoverflow.com/a/28433571/884640)

### DIFF
--- a/examples/asyncio/toasyncgenerator.py
+++ b/examples/asyncio/toasyncgenerator.py
@@ -1,6 +1,6 @@
-import asyncio
-
 import rx
+asyncio = rx.config['asyncio']
+
 from rx.concurrency import AsyncIOScheduler
 from rx.observable import Observable
 from rx.internal import extensionmethod
--- a/examples/autocomplete/autocomplete_asyncio.py
+++ b/examples/autocomplete/autocomplete_asyncio.py
@@ -7,7 +7,8 @@ Uses the RxPY AsyncIOScheduler (Python 3.4 is required)
 """
 
 import os
-import asyncio
+import rx
+asyncio = rx.config['asyncio']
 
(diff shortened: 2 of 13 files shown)


hangtwenty:
BTW, I'm using Python 2.7.5, and have `trollius` installed.

All tests pass :surfer: Also the example `examples/autocomplete/autocomplete_asyncio.py` now works  as long as you [replace `yield from ...` with `yield From(...)`](http://trollius.readthedocs.org/en/latest/using.html). I think we can leave it to users to figure out this example from the README, but if this PR looks good, I'd like to update the README to say `AsyncIOScheduler` _can_ work with Python 2, if you install `trollius`.

Any other testing I can do?

hangtwenty:
Regarding `nonlocal` replacement, would it be preferred to use the idiom found in `asyncioscheduler.py` (single item list)?

```python

        handle = [self.loop.call_soon(interval)]

        def dispose():
            # nonlocal handle
            handle[0].cancel()
```

dbrattli:
This looks great! You are also right about replacing `nonlocal` with the singleton list idiom for py2/py3 compliance. Please update the README, and keep up the good work :-)

PR #39, opened 2015-04-12
added TkinterScheduler.schedule_periodic() so scheduled actions run in t...

pillmuncher:
...he tkinter mainloop instead of a Timer thread.

My tkinter program wouldn't quit properly because of a sleeping Timer thread
from an Observable.interval(TICK, scheduler=TkinterScheduler(window)) stream.
My research showed that TkinterScheduler doesn't overwrite the default implementation of
schedule_periodic() defined in the Scheduler base class, which uses a
threading.Timer. I overwrote this method in TkinterScheduler to run periodic
actions in the tkinter mainloop. Now my program quits properly.

### DIFF
--- a/rx/concurrency/mainloopscheduler/tkinterscheduler.py
+++ b/rx/concurrency/mainloopscheduler/tkinterscheduler.py
@@ -73,3 +73,31 @@ class TkinterScheduler(Scheduler):
 
         duetime = self.to_datetime(duetime)
         return self.schedule_relative(duetime - self.now(), action, state)
+
+    def schedule_periodic(self, period, action, state=None):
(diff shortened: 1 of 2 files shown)


dbrattli:
Good job, thanks a lot! I actually thinks I have experienced this bug myself without finding out what the problem was.

dbrattli:
I'm tempted to remove the timer based `schedule_periodic` from `scheduler.py`since this is a rather nasty bug that could impact the use of other schedulers as well.

pillmuncher:
If you want, I can look into it. Just, so we don't both work on this problem.

dbrattli:
Yes please. That sounds great!

PR #44, opened 2015-04-19
More informative error message

hangtwenty:
Use `raise` instead of `raise <exception>` so that the original exception comes through (more informative -- shows client where their code is causing the problem)

### DIFF
--- a/rx/autodetachobserver.py
+++ b/rx/autodetachobserver.py
@@ -14,9 +14,9 @@ class AutoDetachObserver(AbstractObserver):
     def _next(self, value):
         try:
             self.observer.on_next(value)
-        except Exception as ex:
+        except Exception:
             self.dispose()
-            raise ex
+            raise
 
     def _error(self, exn):
         try:


dbrattli:
So that was the problem. I've been trying to solve this for quite some time but given up. I thought the two lines `raise` vs `raise ex` would do the same thing. But shouldn't we just use a `try ... finally` instead?

hangtwenty:
Well, `try`/`finally` could be relevant (what do you want to do with them?). `raise` without args is very useful! [Documentation is here.](https://docs.python.org/2/tutorial/errors.html#raising-exceptions)

dbrattli:
You are right. It's probably better the way you suggested so I merged your request. Here's btw. the equivalent code from Rx.NET, and it doesn't look any better with an if-test within the `finally` clause.

```c#
            var __noError = false;
            try
            {
                observer.OnNext(value);
                __noError = true;
            }
            finally
            {
                if (!__noError)
                    Dispose();
            }
```

dbrattli:
But I'm so happy for this fix. I've scratched my head so many times about this issue. Thanks a lot!

PR #46, opened 2015-04-30
Improve README header

leinz:
The triple dots in the main header looks weird.


dbrattli:
Nice to see that you are looking at RxPY @leinz . If you find any code that needs to be fixed, then please make the pull request against develop.

leinz:
Ok, will remember that.

PR #49, opened 2015-06-01
Added Qt mainloop scheduler

jdreaver:
I have added a `QtScheduler` to RxPY. It uses a `QTimer`, which is already very flexible, to perform any scheduling tasks. See the `QtScheduler._qtimer_schedule` method to see the meat of the class. It is compatible with PyQt4, PyQt5, and PySide (I have tested all 3 on my machine, and there isn't a reason there should be any fundamental behavior difference between the three).

In order to work, all the scheduler needs is a way to instantiate a `QTimer`. Currently, I have the user pass in the `QtCore` module from whatever Qt wrapper they decide to use. This feels un-Pythonic to me for some reason. I have a couple of alternatives:

1. Allow the value of `qtcore` in the constructor to be `None`, and have us try to predict what framework to use. Basically, we just try to import them in succession until we don't get an `ImportError` (see the top of the new unit test file in this pull request). The problem with this approach is that we might guess incorrectly.
2. Instead of passing in `QtCore`, have the user pass in `QtCore.QTimer`. This way, they could perform dependency injection if they wanted slightly different `QTimer` behavior from their own subclass. However, I am worried that users will get confused and instead pass in an in instantiated `QTimer` instead of simply the class.
3. Somehow allow the user to select the Qt wrapper to use based on an environment variable, or by using the `rx.config` dict. Then, all we do is import whatever framework that have selected. I think this is the way to go, but I would like input on how to do it correctly. I am not sure that `rx.config` is designed to be modified by users.

I also added an example. The unit tests and the example are basically copies of the tkinter versions. I just wanted to make sure the scheduler worked correctly for now without being too creative. In the future I will probably add more examples as I use RxPY more.


dbrattli:
This is fantastic. Great work!

jdreaver:
Thanks!

I can't close the pull request, but since it is merged it will show up as Closed.

dbrattli:
Yes, you are right :-) I think this looks good for now. I guess you need more use of the scheduler to figure out if and how it should be improved. Hopefully I will get the time to play with it this weekend.

jdreaver:
Indeed. I started writing a [TodoMVC implementation](https://github.com/jdreaver/RxPY-TodoMVC) using RxPY and PyQt4. I will just add features to it until I feel I have a good grasp of how to implement a Qt app with RxPY. Then I will try to integrate RxPY with my real applications at work.

PR #52, opened 2015-06-25
schedule_periodic method implementation in AsyncIOScheduler

rrader:
Default implementation uses Timer which creates separate thread. It makes a big problem because asyncio supposed to be in one thread, and synchronization primitives of asyncio are not thread-safe.

Provided implementation uses asyncio event loop capabilities to schedule periodical task.


dbrattli:
Thanks a lot! We should just remove default impl. so other schedulers don't have the same problem.

PR #58, opened 2015-10-03
Make pluck docstring more accurate and add pluck_attr

jdreaver:
The docstring of `pluck` says that it grabs the "value of the specified **property**" from each element, but in reality it is using `__getitem__` on each element. Therefore, I changed the docstring of `pluck` to be more specific, and I added a new `pluck_attr` method that actually grabs properties using `getattr`.

I was going to say that `pluck` uses `__getitem__` to get values, but I think that to a beginner the phrase "dict-like" is easier to understand. I also mention specifically that `pluck` calls `element[key]` in the docstring, in case a user calls `help(Observable.pluck)` and can't see the simple implementation.

### DIFF
--- a/rx/linq/observable/pluck.py
+++ b/rx/linq/observable/pluck.py
@@ -3,14 +3,35 @@ from rx.internal import extensionmethod
 
 
 @extensionmethod(Observable)
-def pluck(self, property):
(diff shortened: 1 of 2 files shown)


dbrattli:
This looks good. Thanks!

PR #59, opened 2015-10-04
Use shields.io to uniformize and fix your badges

DavidJFelix:


dbrattli:
Great! Thanks a lot!

DavidJFelix:
Just a quick note about this: github camo (their image proxy) sometimes makes (any variety of) status badges behave funny. The one that seems to have the most trouble is travis, since it changes the most regularly. It seems that camo has a few factors with caching. Visiting the image directly (via github camo link, not README link) will cause camo to refresh the content, but even a force refresh within the README will not. This is possibly a referrer URL timeout.

PR #63, opened 2015-11-04
Terminate Observable.from_iterable on dispose

### LINKED ISSUE #62: Possible bug when disconnecting subscriber from iterable
I try to use iterator as observable but I want this observable to be as cold as possible, i.e. generate new values only when there is any subscriber. But when I dispose subscriber iterator still works. The code is as follow:

``` python
from time import sleep
from rx import Observable
from rx.concurrency import new_thread_scheduler


def myiterator():
    for i in range(100):
       sleep(1)
       print('Publishing value: {0}'.format(i))
       yield i

source = Observable.from_(myiterator(), scheduler=new_thread_scheduler)
subscription = source.subscribe(
    lambda v: print("Value published: {0}".format(v)),
    lambda e: print("Error! {0}".format(e)),
    lambda: print("Completed!")
)

sleep(3)
print('Before dispose')
subscription.dispose()
print('After dispose')
sleep(3)
print('Finish')
```

and the output is:

```
Publishing value: 0
Value published: 0
Publishing value: 1
Value published: 1
Before dispose
After dispose
Publishing value: 2
Publishing value: 3
Publishing value: 4
Finish
```

I suppose there should be no publishing value after subscriber disposal. Am I correct?

jdreaver:
This fixes #62.

I couldn't think of a convenient way to add a test for this short of the example that @mikus provided. I tried creating an iterable by first creating an observable with the test scheduler and then calling `to_iterable()`, but that doesn't work (I didn't expect it to).

### DIFF
--- a/rx/linq/observable/fromiterable.py
+++ b/rx/linq/observable/fromiterable.py
@@ -1,6 +1,7 @@
 from rx.observable import Observable
 from rx.anonymousobservable import AnonymousObservable
 from rx.concurrency import current_thread_scheduler
+from rx.disposables import Disposable, CompositeDisposable
 from rx.internal import extensionclassmethod
 
 
@@ -26,8 +27,11 @@ def from_iterable(cls, iterable, scheduler=None):
 
     def subscribe(observer):
         iterator = iter(iterable)
+        is_disposed = [False]
 
         def action(action1, state=None):
+            if is_disposed[0]:
+                return
             try:
                 item = next(iterator)
             except StopIteration:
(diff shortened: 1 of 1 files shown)


dbrattli:
Thanks for the patch. I'm currently a bit unsure if it's `from_iterable()`or the `NewThreadScheduler` in the case of `schedule_recursive()` that needs to be fixed. Recursive scheduling probably don't make sense for schedulers such as the new thread scheduler. I will look more into it this weekend. It would be nice if we didn't use `schedule_recursive()` but I'm not sure right now to compose the disposable side of things.

jdreaver:
No problem! If you find a better solution, then feel free to close this.

dbrattli:
I'm want to see if I can remove `schedule_recursive` but then I will need your dispose logic anyway, so I'm merging your patch and working for that instead.

jdreaver:
:+1:

PR #68, opened 2015-11-30
Removed Microsoft visual studio solution files.

droxer:


dbrattli:
Please explain why you want to remove the VS solution/project files

droxer:
@dbrattli I don't think for a Python project to depends on specified IDE is kind of good practice.

dbrattli:
@droxer, RxPY does not depend on these files, but Visual Studio depends on the solution files if you want to use RxPY with Visual Studio. Thus you are free to add an ignore rule for them in your favorite editor.

PR #74, opened 2016-01-22
Adding Eventlet driven event loop

jalandip:


dbrattli:
This is great work. Note that you also want to override `schedule_periodic()` since the default implementationi in scheduler.py will call `Timer()` which uses a thread and will give you weird behaviour. Look at `AsyncIOScheduler` or `TkinterScheduler` for examples on how to implement it.

PR #77, opened 2016-02-20
Scheduler for use with wxPython applications

bosonogi:
Ported from `QtScheduler` and the `timeflies` example.


bosonogi:
Hi, I came here after listening to the [software engineering daily podcast](http://softwareengineeringdaily.com/2016/02/15/reactive-programming-with-matthew-podwysocki/) and hope to use rx in the future.

Is there a better way to go about canceling all scheduled actions than the explicit scheduler method call I added? Some kind of a dispose call somewhere perhaps?

dbrattli:
Great work! Looks like they use the methods `start()` and `shutdown()` in [RxJava](https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/schedulers/Schedulers.java) so you can try to align with that.

PR #89, opened 2016-04-25
Rename Subject to Stream

dbrattli:
Subjects are confusing for most developers. But they are really just both an observer and an observable. An Observable is a stream producer, and an observer is a stream consumer. A Subject is both a consumer of events and a producer of events, so why don't we call it a Stream? 

Main Rx documentation use the word "stream" without telling what it really is:
- Easily create event streams or data streams.
- Compose and transform streams with query-like operators.
- Subscribe to any observable stream to perform side effects.

But there are no Stream artifacts in Rx. Streams are just some abstract concept. If we renamed Subject to a Stream, then that would make the on-boarding a lot easier for new developers using Rx.

That one should be careful with the use of Subjects is a completely different story. But Python is an imperative programming language, and we do not usually warn developers against using Lists even if they are mutable. Why should we then warn against the use of Subjects?

Let's call the stream for a Stream!


frederikaalund:
Indeed, Stream is a more intuitive name than Subject. I completely agree that it is strange that streams are merely concepts and not artifacts in Rx.

However, I do not think renaming Subject to Stream is a good idea. The biggest con is that this change would deviate from the ReactiveX "specification" (I know that there is no official spec). Still, the name Subject has become a de-facto standard in all Rx implementations.

One of the greatest benefits of Rx is that it is cross-platform. Personally, I use both RxJS and RxPY and I love the fact that the code is almost identical. It makes it easy to port ideas from one language to the other. I know that `bufferWithCount` becomes `buffer_with_count` and vice versa, but that is simple mapping. The same does not go for `Subject` and `Stream` People would at first be confused of this strange new `Stream` artifact. Let's avoid that confusion.

Case in point: `on_error_resume_next`. This function has a somewhat misleading name. Still, the name sticks because it has become a de-facto standard. The [official Rx documentation](http://reactivex.io/documentation/operators/catch.html) even points this out!

> This implementation borrows the confusing nomenclature from Rx.NET, in which `onErrorResumeNext` switches to a back-up Observable both on an error *and* on a normal, error-free termination of the source Observable.

Renaming Subject to Stream is a good idea for a Rx 2.0 "specification". Let [them](https://github.com/orgs/ReactiveX/people) know! However, let's keep things standard for the current iteration of Rx.

Currently, RxPY is already deviating from the de-facto nomenclature with `catch_exception` which is more commonly known as `catch_error`. If anything, an effort should be made to fix such deviations in milestone 2.0.

This is just my personal opinion. Thanks for all your great work on RxPY!

PR #99, opened 2016-05-29
Update Getting Started.ipynb

MajorTal:
spelling


dbrattli:
Looks like I've been writing German. Thanks for the fix.

PR #101, opened 2016-06-12
Sphinx Docs, Logo and Updated Docstrings (partial)

ninmesara:
- Added logo
- Added sphinx docs (following  the structure of the official ReactiveX docs)
- Placed marble diagrams in the operator pages using(e.g. [all](https://ninmesara.github.io/RxPY/api/operators/all.html), [buffer](https://ninmesara.github.io/RxPY/api/operators/buffer.html)). The marble diagrams were placed using a script that tried to guess the correct operator from the filename, and there may be mistakes.
- Updated some docstrings

Keep in mind that the docs are just a sketch, and they were mostly automatically generated.
I hope this can serve as a useful scaffold to build on.
Even updating the docstrings so that they are valid ReST would improve the docs a lot.

You can find the HTML docs [here](https://ninmesara.github.io/RxPY/).

The marble diagrams are included, which makes the repository very large, but it doesn't seem right to link them from external websites.


ninmesara:
Some operators are not properly grouped yet - I'll look into it soon.
The goal is to follow ReactiveX's structure (group `combine_latest` with `with_latest_from`, for example)

ninmesara:
[Here](https://ninmesara.github.io/RxPY/api/operators/combine_latest.html) is an example of a page with interactive marble diagrams.

dbrattli:
Hi @ninmesara. This looks great. Really nice work! Not sure I want this in the master branch right now since it's a sketch and still quite unfinished. Perhaps we should create a separate feature/doc branch that you can work on until it is ready enough to be merged and published? It doesn't have to be complete in order to be ready, but at least some parts of the doc needs to be usable and in the format we want. However, I'm afraid I have more than enough to do on the book project right now, so I hope that you will have the time to keep working on the docs!?

A few comments:

- "Under construction" should be removed from all pages. Haven't seen that since the 1990s. Everything is under construction, always. If it's not ready to be published, then we keep it in a feature branch. When it's ready to be published we merge it to develop. http://jeffkreeftmeijer.com/2010/why-arent-you-using-git-flow/
- I don't see ReactiveX as an external website. It's another repo, but both projects are under github.com/ReactiveX. So I don't think we should copy the images, especially since they are so large, and we may not want to use all of them either.
- Also I don't want the copyright to myself. Especially since we are using text, structure, images and code from ReactiveX. Use similar license as ReactiveX. "ReactiveX is a collection of open source projects. The content of this page is licensed under Creative Commons Attribution 3.0 License, and code samples are licensed under the Apache License".
- Btw also started working on a logo, but we don't need do decide on that now. https://twitter.com/dbrattli 
- Also look at RxJS doc structure. https://github.com/Reactive-Extensions/RxJS/tree/master/doc

But super cool that you have started doing this work 😄 If you are ok with it, I will create a separate feature/doc branch that you can work on until we are ready to merge it to develop and eventually to master. Does this sound like a plan? You can also get write access to the repo to make things easier, but let's do a few pull requests first. Docstring changes for the operators is something we can accept for the develop branch so the doc branch does not have to touch .py files in order to avoid merge conflicts.

Dag

ninmesara:
> I hope that you will have the time to keep working on the docs!?

Not much time, but eventually I hope to:

* Change the docstrings into valid ReST
* Add interactive marble diagrams to all operators in which it is feasible (like these [here](https://ninmesara.github.io/RxPY/api/operators/combine_latest.html) and [here](https://ninmesara.github.io/RxPY/api/operators/merge.html)). I have the javascript code written for some operators already, and it is just a question of "instanciating" the code with specific observables, and different functions (like sum, string concatenation, etc). The goal would be to at least replicate [these diagrams](http://rxmarbles.com/) (which are slow, and won't work on mobile), with more functions and datatypes (strings instead of numbers, for example).
* Add some examples

This is all very easy, but tedious. I can't give you a timeframe, though.

> "Under construction" should be removed from all pages.

Ok, I agree. I've put the "under construction" banner in the pages that were 100% automatically generated from the docstrings and ReactiveX images without any kind of error checking or manual intervention, but since this is not something to release right away, it makes sense to remove them all.

>  I don't think we should copy the images, especially since they are so large, and we may not want to use all of them either.

Here I disagree. First, we might want to edit the images to convert the operator names from `camelCase` into `snake_case` (not a priority right now, but something we might want to do in the future). Second, having the images hostd online makes it impossible to preview the docs offline, which is undesirable. Third, the download price is paid only once, since the images will change rarely if ever (see above). The only strong argument against is the fact that it might make the Python package too big for end users. Is there any way to distribute the package without the docs or with the docs but without the images?

> Also I don't want the copyright to myself.

Sure, I'll change this.

> Btw also started working on a logo

I have another version of the logo (in attachment) that is more similar to yours:

other-logo

It is ReactiveX's dragon (side fin included) and Python's colors. I'm not sure which one I prefer (but I prefer either of mine to yours). My repo has vector traces of ReactiveX's logo (in SVG format) so you can play with it if you want.

> Also look at RxJS doc structure

I've looked at it too (my first version of the docs had one page per operator, too), and it's a great source for examples, but I prefer ReactiveX's structure, as they group related operators together.
Maybe I should include include links for the corresponding RxJS page for each operator, at least until we have enough examples in Python. I already include links to ReactiveX's website.

> If you are ok with it, I will create a separate feature/doc branch that you can work on until we are ready to merge it to develop and eventually to master. [...] Docstring changes for the operators is something we can accept for the develop branch.

I don't know if I like this idea. In order to build the docs, Sphinx needs to have access to the python source code, and it makes sense to have the documentation and code in the same branch. Or can sphinx get the source code from a different branch. Of course, you could do a `setup.py install` in the *develop* branch different branch and switch to the *features/doc* branch, but it doesn't seem like a sane workflow to me. Also, I don't see how this could lead to merge conflics.

dbrattli:
Docs will not be built into the distributed package/wheel, so it's really not a big deal if we copy the images.

With a separate branch you get everything, both code and docs. The reasons for checking docstrings into develop is that you can merge develop into the feature branch, so you have the same thing in both branches. If you decide to change docstrings in the feature branch you will get conflicts every time someone changes the docstring in develop. Your call.

Feel free to make any PR against [feature/docs](https://github.com/ReactiveX/RxPY/tree/feature/docs). You can always continue to develop in your private branch to be more agile, and make a new PRs when you have something you want to share.

ninmesara:
Ok, I get it. Seems great, and as long as merge into docs, it will work well. Sorry for my inexperience with git.

PR #102, opened 2016-06-15
Update README.md

wenjianhn:
We do support Eventlet.


prokaktus:
Also I suggest to remove downloads badge.
badge

Downloads count broken on Pypi and badge with `downloads 41/month` looks... not very optimistic, at least.

dbrattli:
Merged into develop

PR #104, opened 2016-06-27
README: link to reactivex.io docs on operators

hangtwenty:
in README...
- Link to the helpful [reactivex.io](http://reactivex.io/) docs re: operators -- such a handy reference to have around, a newcomer might not know about it. Now... most devs are probably coming here _from_ reactivex.io, but what if they aren't? Just want to make sure they have the docs in hand.
- Also link to the 'passing unit tests' mentioned, because why not :) it's something to be proud of!

PR #105, opened 2016-06-27
README: links to more resources on ReactiveX

hangtwenty:
Same idea as #104, but adding more links to the reactivex.io resources.


dbrattli:
Merged into develop

PR #116, opened 2016-08-17
Make AsyncIOScheduler.now use milliseconds

JobNoorman:
It incorrectly used the result from asyncio's AbstractEventLoop.time() method
which returns the current time in seconds. This caused, for example,
Observable.delay() to misbehave when using the AsyncIOScheduler.


dbrattli:
Thanks a lot for discovering and fixing this

PR #118, opened 2016-08-22
Don't override custom exceptions in throw.py

frederikaalund:
Say I have a custom exception:

    class TimeoutException(Exception):
        pass

I throw this if some `observable` times out using

    observable.timeout(1000, Observable.throw(TimeoutError))

so that I can later switch on the type of the exception in a `.catch_exception` call.

The file change allows this behaviour. The old behaviour was to override the exception type to simply Exception (since `type(TimeoutException) != Exception`).

### DIFF
--- a/rx/linq/observable/throw.py
+++ b/rx/linq/observable/throw.py
@@ -24,7 +24,7 @@ def throw(cls, exception, scheduler=None):
 
     scheduler = scheduler or immediate_scheduler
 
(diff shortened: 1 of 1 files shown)


diorcety:
Hi,
I'am agree with you.

dbrattli:
I also agree, but all pull requests to develop branch please.

frederikaalund:
Created a new pull request targeting the `develop` branch in #119.

Issue #115, opened 2016-08-16
Use multiprocessing Condition instead of threading Condition


diorcety:
It is the same, but multiprocessing handles signal interruption. Which is usefull to interrupt a program blocked in blocking observable

dbrattli:
Still had problems with `multiprocessing` on OS X (too many open files) when running unit-tests. Have gone back to `threading` module as default, but you can now use `multiprocessing` by setting the concurrency option in the config:

```python
import multiprocessing
from rx import config
config["concurrency"] = multiprocessing
```

RoryKelly:
Any more info on the "too many open files" with multiprocessing? I'm having the same problem when processing large amounts of audio files on OSX.

thomasnield:
@RoryKelly without knowing the full context of this, have you tried using operators to limit how many files get processed at at time, like `buffer()` and then `flat_map()`?  I'd try to constrain that pipeline to limit your paraellelism, if I understand what's going on correctly.

RoryKelly:
I had too many threads opening too many large files at the same time. Apologies im coming from Rxjava and my main problem is understanding the relationship between the GIL, ThreadPoolScheduler and "config["concurrency"] = multiprocessing"

I want to run multiple observable streams in parallel as fast as possible, should I 

1. use the technique outlined [here](https://github.com/ReactiveX/RxPY/blob/master/examples/parallel/timer.py) 
2. Use ThreadPoolScheduler and subscribe on.
3.  Use ThreadPoolScheduler and subscribe on with "config["concurrency"] = multiprocessing" (does this even have an effect? I'm imagining that this transparently swaps from threading to process based "threading". https://docs.python.org/2/library/multiprocessing.html though I feel that's to optimistic)

MainRo:
Closing due to inactivity. Please re-open if needed.

PR #120, opened 2016-08-23
Use multithreading for Condition instead of threading in order to han…

diorcety:
…dle signal interruptions
cf #115

### DIFF
--- a/rx/concurrency/eventloopscheduler.py
+++ b/rx/concurrency/eventloopscheduler.py
@@ -1,4 +1,5 @@
 import logging
+import multiprocessing
 import threading
 from threading import Timer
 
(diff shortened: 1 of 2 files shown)


dbrattli:
Thanks. I'm going to have a look at this PR this weekend. Just want to check that there is no perf degradation. We might need a wrapper for it in the config similar to `Lock`, but I can take care of that.

diorcety:
?

dbrattli:
Had to revert since it broke develop. https://travis-ci.org/ReactiveX/RxPY/builds/155705742

diorcety:
Strange, Travis on my PR was good

PR #125, opened 2016-09-01
catch_exception: Pass the source observable on

frederikaalund:
Allows the error handler to retry the source observable. Similar to the functionality provided by [RxJS 5's `catch`](https://github.com/ReactiveX/rxjs/blob/7b23e88db21302a10c2a163952b8da197d18cbfe/src/operator/catch.ts#L18).

Please refer to ReactiveX/rxjs#141 for the rationale.


frederikaalund:
Some of the tests fail due to the new interface. It should be trivial to fix.

PR #130, opened 2016-10-24
 Add a scheduler for the GObject/Gtk main loop

fqxp:
This work is closely based on the Qt main loop scheduler. Seems to work fine so far, tests are included, but will not automatically be run be travis-ci.

I'm open for discussing this code :)


dbrattli:
This looks great. Thanks!

PR #132, opened 2016-10-30
Fix GLib import in GtkScheduler

fqxp:
GLib was imported in the wrong place which made the GtkScheduler not work.


dbrattli:
Thanks for fixing this

PR #140, opened 2016-12-06
Fix Tornado Typo

rgbkrk:
Same as #139, didn't realize you had a `develop` branch.


dbrattli:
Thanks, it will eventually bubble over to master.

Issue #133, opened 2016-11-20
`group_by()` doesn't work with Reducing Operators


thomasnield:
I was working on some course material and then I noticed that using `group_by()` is not working with the `count()` operator. 

For instance, if I emit a sequence of letters and count their occurrences using `group_by()`, the counts are all short by 1. 

```python
from rx import Observable

Observable.from_("alphabetagammadeltaepsilon") \
    .group_by(lambda s: s) \
    .map(lambda grp: grp.count().map(lambda ct: "{0}-{1}".format(grp.key,ct))) \
    .merge_all() \
    .subscribe(print)
```

**OUTPUT:**

```
a-5
l-2
p-1
h-0
b-0
e-2
t-1
g-0
m-0
d-0
s-0
i-0
o-0
n-0
```

Even more unsettling is sometimes the counts just come out to zero when several occurrences do exist. 

```python
from rx import Observable

Observable.from_(["alpha", "apple", "beta", "bat", "gamma"]) \
    .group_by(lambda s: s[0]) \
    .map(lambda grp: grp.count().map(lambda ct: "{0}-{1}".format(grp.key, ct))) \
    .merge_all() \
    .subscribe(print)
```

**OUTPUT:**

```
a-0
b-0
g-0
```

I have confirmed RxJava does not have this problem. Interestingly though, not doing any reducing like `count()` or `to_list()` will result in no emissions being missed.

```python
Observable.from_(["alpha", "apple", "beta", "bat", "gamma"]) \
    .group_by(lambda s: s[0]) \
    .merge_all() \
    .subscribe(print)
```

**OUTPUT:**

```
alpha
apple
beta
bat
gamma
```

Is there something I am missing? Or is this a bug?

thomasnield:
Interestingly I did not have this problem on my Linux machine last night. It might have been running an older version of RxPy.

thomasnield:
Just rolled back to RxPy 1.5.2. Now it is working fine. It looks like 1.5.3 is broken.

thomasnield:
This regression still exists with 1.5.4. For the given code below:

```python
from rx import Observable

Observable.from_("alphabetagammadeltaepsilon") \
    .group_by(lambda s: s) \
    .map(lambda grp: grp.count().map(lambda ct: "{0}-{1}".format(grp.key,ct))) \
    .merge_all() \
    .subscribe(print)
```

This is the output when ran with 1.5.2

```
a-6
l-3
p-2
h-1
b-1
e-3
[...]

PR #150, opened 2016-12-21
Fix recursive scheduling in concat.

kstreee:
I encountered the maximum recursion exception while using following combinations.

```
def creator(observer):
    observer.on_next(1)
    observer.on_next(2)
    observer.on_completed()

Observable.while_do(lambda _: True, Observable.create(creator)).subscribe(print)
Observable.create(creator).repeat().subscribe(print)
```

The `repeat` and `while_do` with `create` leads maximum recursion exception because of `concat`.
So I searched about some issues which are related to this issue, and I found this #133 . To resolve this issue and the #133, I changed a scheduler that `concat` is using from `immediate_scheduler` into `CurrentThreadScheduler()`, and I tested that this change resolve both issues.  But, I'm not sure that using a new object by calling `CurrentThreadScheduler()` is okay to `concat`, so please check this out. (F.Y.I. former version of the issue #133, `concat` used `current_thread_scheduler` as a scheduler, and in that case, there wasn't maximum recursion exception.)

### DIFF
--- a/rx/concurrency/currentthreadscheduler.py
+++ b/rx/concurrency/currentthreadscheduler.py
@@ -20,12 +20,12 @@ class Trampoline(object):
         while len(queue):
             item = queue.dequeue()
             if not item.is_cancelled():
-                diff = item.duetime - current_thread_scheduler.now
+                diff = item.duetime - item.scheduler.now
                 while diff > timedelta(0):
                     seconds = diff.seconds + diff.microseconds / 1E6 + diff.days * 86400
                     log.warning("Do not schedule blocking work!")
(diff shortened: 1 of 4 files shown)


kstreee:
There was a mistake of choosing a base branch. I changed the base branch from master to develop. I'm sorry about that.

kstreee:
I wrote test cases for both issues (#133, and this). Please review the pull request, and it would be a great pleasure for me. Thanks.

dbrattli:
I was worried about #133, so really great that you made a test for that case as well. Nice work!

Issue #149, opened 2016-12-20
How to contribute to RxPY


AXGKl:
Hi Dag, 

this is a "meta" question. If there was any comm. thread in the past clarifying those things then link me to it, did not find sth.

Problem Statement:

- Github, until you, as owner, merge or reject a PR, does not allow separate PRs, at least it seems so. 
- Any commit into my fork does land in the one open PR on development branch (just closed [this](https://github.com/ReactiveX/RxPY/pull/148) since it is already containing 3 unrelated small things and I have a few more locally still)
- Since this is a (fantastic) **spare time** project of yours, you can't check open PRs on a daily basis. 

My problem is that I have not this one big contributuion going on, justifying (in my world) a feature branch. But, as a beginner with RxPY, I have many smallish things, which do not justify branches (or do they?).

So, how you want contributions in the following cases:

1. Small corrections like typos in the docstrings
1. Small fixes like the missing default for scheduler in `time_interval`
1. Additional Documentation like [this](https://github.com/axiros/RxPY/tree/develop/notebooks/reactivex.io)
1. Additional tests

Alternatives as far as I can see:

1. Letting all accumulate in one PR (with clear commit messages) and you accept, reject or cherry pick - easy for us but maybe you don't like that(?). We could redeclare/generalize the PR title with any new addition.

1. Separate branches for each and anything

1. Contributors wait with next push to their forks until you accept their current open PR on dev. branch

1. Other flows I'm not aware of?

-----

Regarding the additional tests: You have tons of tests based on the Virtual Time Testscheduler, allowing fast test times. But they are hard to understand for beginners. Would you even appreciate some tests based on the normal schedulers, e.g. using marbles to set up streams? 

Upside: Tests will better document, for beginners
Downside: Testtime

My suggestion: Putting those beginner use cases in IPython Notebooks, incl assertions. I could add assertions e.g. [here](https://github.com/axiros/RxPY/tree/develop/notebooks/reactivex.io).
Not have them run on travis all the time but on demand. Should one fail then beginners will send issues anyway.

-----

Cheers, many thanks for this work. Python has many async features but this is as far as I can see the only project which allows for proper coordination.

dbrattli:
[...]

PR #152, opened 2016-12-22
RXPy Beginner Documentation Package

AXGKl:


AXGKl:
Hi Dag,

*first of all: sorry for the big PR, reject and feedback how you want it if you want it split.*

It is all about lowering the entry barrier for noobs like me on reactive programming in python using your RXPy. It is an ipython notebook based translation of ALL(!?) the supported operators in http://reactivex.io/documentation/operators.html into working code, beginners can play with.

- Along the way I realised here and there smallish docu or signature problems and corrected them, thats why the PR has so many commits.

- Also futures in Py2/trollius did actually work, so I thought it makes sense to have it tested on travis.

Why?

I guess in your world its totally useless to create python code from the js examples on reactive.io but for a beginner I think it does help - for myself the one notebook you did have was really great to start (plus you seem to not like writing books ;-)
Thats why I think it makes sense to include the notebooks.

I also intend now to create other notebooks with practical examples of RX as a tool to solve real use cases like FIFOS, caches, connection pools - but that with time.


Let me know your thoughts.

- If you don't like it I can also put it on a repo on our site.
- If you like it a lot I can also set up some docker stuffs to have a live version in the Internet.

Cheers,
Gunther

AXGKl:
Hi, regarding https://github.com/ReactiveX/RxPY/issues/149#issuecomment-268822995, maybe I can save you time by explaining a bit more:

### 1. tofuture

Affected: Unittests only (the last commits)

For the notebooks `tofuture`, I did a `pip import trollius` since I'm old and think still Py2 > Py3 but thats a different story. In any case I realized the testsuite breaking after trollius install.  
Reason was that you check for availability of futures via `rx.config` which says then yes.  
Problem: A few tests use Py3's async language addition -> test not skipped and breaks at module import time with syntax error.

=> My suggested solution in the PR is to break appart the future related tests into 2 modules:
 
- 1 which works in Py2 as well when you have trollius or tornado's futures, skipping mechanics via rx.config
- 1 which works only in Py3 since using `async`


### buffer openings

Affected: https://github.com/ReactiveX/RxPY/pull/152/commits/2eb39d876fbb15eac60c5086539db81ce9b4b7b5

**This one you should carefully check.** I'm a beginner with the framework but I strongly believe it must be like that. 
A notebook use case was then working and before not.

### Marble Strings with a Bool Cast Convention

Affected: https://github.com/ReactiveX/RxPY/pull/152/commits/a1c667f76f053418d86772c4bab572693b5317d3

This one you can safely skip if you don't like the API break - its a suggestion and I found it very useful during testing of a few operators to directly get booleans in a control stream, which e.g. define window closings.

Not in use anywhere, did refrain to use it in the notebooks. 


### The marbles tests

Affected: https://github.com/ReactiveX/RxPY/pull/152/commits/b8876957fae799750491151538a53fe238ab6fd8

I think it adds too much test time for not much benefit.

We should go back to the version before, it did work after your fix with the addition few milliseconds (I was in panic mode seeing intermittent nose failures and wanted to do it right. After the commit I saw your fix).




### The notebooks themselves.

Affected: Only files [here](https://github.com/axiros/RxPY/tree/develop/notebooks/reactivex.io)

They are not considered finished, I want to add a few gui features like widgets with sliders and nicer logging in html but:

They are pretty complete, regarding operators, meaning: Considered by me useful at this stage and consistent.

Still, remembering Stalz' comment about Readme.md: This is a lot about style and it might not fit at all with yours. => I have no problem to craete a repo in my company's space and maybe even a live version which you could refer to in docu. An alternative would be a contrib repo here or a contrib directory (?)




----

The rest should be pretty easy to understand.


Cheers, 

Gunther

dbrattli:
I really like this. Great work!

AXGKl:
wow ...

thats really a big honour.

PR #158, opened 2017-01-14
fixed issue-157: `generate()` fails to re-subscribe

yarosla:
1. Added test case that was failing
2. Fixed `generate()` method
3. Fixed test in `to_dict` that fails under Python 3.6 (due to dicts are now ordered)


dbrattli:
Thanks for fixing this.

PR #159, opened 2017-01-19
The only way that the GTK Demo would work for me.

UmlautBioEye:
Added an additional import statement at the top of the file. This appears to be needed in addition to the import statement which appears __init__(self). It is the only way that the demo [time flies] program would run on my system.
-UmlautBioEye
-19 Jan 2017


UmlautBioEye:
Implemented Code which fqxp came up with four months ago.
It built successfully in Travis.
Will try it at home and if it works, I will withdraw my Pull Request.
-19 Jan 2017
-12:42 pm CST

PR #176, opened 2017-04-19
Fix calling to_timedelta twice

MikeSchneeberger:
In EventLoopScheduler.schedule_periodic the 'to_timedelta' method is called twice: Once in EventLoopScheduler.schedule_periodic and once in EventLoopScheduler.schedule_relative.


dbrattli:
Thanks for the fix!

PR #180, opened 2017-06-13
Align buffer_with_time_or_count signature with doc

ikalnytskyi:
According to docs, `buffer_with_time_or_count` has an optional scheduler
parameter but in reality it's mandatory. Let's make it optional for real
as passing `None` as third argument all the time is a bit inconvenient.


ikalnytskyi:
@dbrattli hey! any change to get it merged?

dbrattli:
Thanks for the PR, and sorry for the delay.

PR #190, opened 2017-07-19
Better error handling for from_callable

mako-taco:
We're currently passing the type of `Exception` to `on_error`, instead of the instance of the exception that was caught.


mako-taco:
sorry, didn't mean to do this against master

PR #192, opened 2017-07-20
Better error handling for on_callable

mako-taco:
We're currently passing the type of `Exception` to `on_error`, instead of the instance of the exception that was caught.


thomasnield:
Good catch. My bad.

mako-taco:
it's possible that we should be saving a traceback when we do this so things aren't impossible to debug when the callable throws, too.

thomasnield:
If you have a helpful way to do that, please put in another PR and we'll take a look! 

Sorry progress has been slow this summer. Dag and I have both been busy.

mako-taco:
Np, thanks for the wonderful port. Another PR will be there soon.

mako-taco:
Np, thanks for the wonderful port. Another PR will be there soon.

PR #201, opened 2017-09-14
Guard against IndexError when peeking empty queue

MaxJohansen:


dbrattli:
Thanks for doing this. Just got the same error myself in a project. Could you please fix the test as well since CI is breaking (see above), and make the pull against the develop branch.

MaxJohansen:
Yeah, I cloned that project you mentioned and decided to fix this bug before I got started. The test passes now, but it DOES change the API of the PriorityQueue. Anyone relying on the IndexError might be in for a surprise.

dbrattli:
Yes, you are right that it changes the API but it doesn't matter since it's internal. But I just realized that you then cannot see the difference between an empty queue and a queue fill with `None`s. I think we should throw an error after all. Could you change it to instead `try/except IndexError` around the lookup and thow it's  own `InvalidOperationException("Queue is empty.") that you can add to `internal/exceptions.py`.

dbrattli:
Thanks!

PR #205, opened 2017-09-27
Improved error raising for adapt_call in internal.utils module

couven92:
In the case the user supplied function raises a `TypeError`, the `func_wrapped` just swallows it and the original exception is lost.

This captures the generated exceptions instead, and throws a new `TypeError` exception with all previuosly captured `TypeErrors` instead.

### DIFF
--- a/rx/internal/utils.py
+++ b/rx/internal/utils.py
@@ -28,18 +28,22 @@ def adapt_call(func):
         if cached[0]:
             return cached[0](*args, **kw)
 
+        exceptions = []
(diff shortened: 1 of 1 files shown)


dbrattli:
This is great. I just tried to help someone with code that throwed in the select mapper, and the exception message was "TypeError: <lambda>() takes 1 positional argument but 2 were given" which is really confusing. I'm considering going back to function inspection again, but let's try your patch first.

couven92:
Yes, had the exact same problem myself :)

PR #210, opened 2017-10-22
`catch_exception`: pass source observable to handler

neilvyas:
Fixes #125, and also some bonus test refactoring! (you can cherry-pick the first commit if you don't want the refactoring.)

re: changes to rx.testing.recorded: I don't really know where to put these functions.
re: docstrings: happy to change formats or whatever.

tests pass locally (py2.7) and I left in a commit with both the "old" tests and the "new" tests to show that I'm not breaking / reducing coverage.

### DIFF
--- a/rx/linq/observable/catch.py
+++ b/rx/linq/observable/catch.py
@@ -4,9 +4,13 @@ from rx.disposables import SingleAssignmentDisposable, \
 from rx.concurrency import current_thread_scheduler
 from rx.internal import Enumerable
 from rx.internal import extensionmethod, extensionclassmethod
+from rx.internal.utils import adapt_call
 
 
 def catch_handler(source, handler):
+    # For compatibility with legacy handlers that don't take ``source``.
+    adapted_handler = adapt_call(handler)
+
     def subscribe(observer):
         d1 = SingleAssignmentDisposable()
         subscription = SerialDisposable()
@@ -15,7 +19,7 @@ def catch_handler(source, handler):
 
         def on_error(exception):
             try:
-                result = handler(exception, source)
+                result = adapted_handler(exception, source)
             except Exception as ex:
                 observer.on_error(ex)
                 return
--- a/rx/linq/observable/catch.py
+++ b/rx/linq/observable/catch.py
@@ -15,7 +15,7 @@ def catch_handler(source, handler):
 
         def on_error(exception):
             try:
-                result = handler(exception)
+                result = handler(exception, source)
             except Exception as ex:
                 observer.on_error(ex)
                 return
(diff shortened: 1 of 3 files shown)


neilvyas:
`py3` fails because I need `from functools import reduce`. Not sure how you guys do version compatibility so I'll just leave it for yall to comment on (also, do yall use `tox`? how do we test multiple versions? I just ran nosetests with a custom 2.7 venv).

tests/test_observable/test_catch.py
```diff
@@ -14,112 +15,119 @@
...
+                return Observable.never()
+            else:
+                return scheduler.create_hot_observable(ms)
+
+        os = [_ms_dispatch(ms) for ms in msgs]
 
         def create():
-            return o1.catch_exception(o2)
+            if use_chaining_api:
+                return reduce(lambda o, catcher: o.catch_exception(catcher), os)
```
mannygit: RxPy is compatible with Python 2.7 and 3.4 so its ok to import reduce from functools.

rx/linq/observable/catch.py
```diff
@@ -15,7 +15,7 @@ def subscribe(observer):
 
         def on_error(exception):
             try:
-                result = handler(exception)
+                result = handler(exception, source)
```
mannygit: Instead of a backwards incompatible change here consider using the [`adapt_call`](https://github.com/ReactiveX/RxPY/blob/master/rx/internal/utils.py#L14) utility function.

This will allow existing code to work with the singe argument form effectively ignoring the source. New handlers written against this feature can accept two arguments and receive both the exception and the source.

neilvyas: thanks for the suggestion!

neilvyas: done, with coverage.

tests/test_observable/test_catch.py
neilvyas: done.

dbrattli:
Thanks for the PR, and sorry for the delay. I guess it's not that good to put too much stuff inside the same PR. The refactoring in particular looks like over-engineering (sorry). There are many `if-else` for the different cases that makes this very hard to read. I don't mind duplicated code in tests, since you can easily fix one test without suddenly breaking all the others.

The other thing is that I want to get rid of `adapt_call()` in the code, so we will need to find another way to do this. Either we need 2 different operators, or have named parameters with different names for the different handlers, or we need to wait for RxPY 2 to make a breaking change.

neilvyas:
I agree that it was too complicated as written, so I deleted all the logic that attempted to infer the expected messages.

PR #213, opened 2017-10-25
still better error handling for `adapt_call`

neilvyas:
This is another go at the problem solved by #205. 

I think this approach does a better job than the previous one because it will preserve the traceback around `TypeError`s raised within the body of an adapted function.

I also added some coverage for error handling. The code is pretty thoroughly documented with explanations since error handling (properly) in python is tricky!

Without better error handling, I'm uncomfortable using this for production code, as some errors would be swallowed (pre-205), or lack the context necessary for debugging and escalation (pre-this PR). Since `adapt_call` is used frequently, this could have posed a significant problem.

### DIFF
--- a/rx/internal/utils.py
+++ b/rx/internal/utils.py
@@ -1,3 +1,5 @@
+import sys
+
 from rx import AnonymousObservable
 from rx.disposables import CompositeDisposable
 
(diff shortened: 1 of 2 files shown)


neilvyas:
oh and [here's some context about re-raising errors and preserving context](https://blog.ionelmc.ro/2014/08/03/the-most-underrated-feature-in-python-3/#id2)

dbrattli:
Thanks for taking another look at this problem. For the next major version of RxPY I want to get rid of `adapt_call`  altogether by splitting operators like `map()` into `map()` and `map_indexed` to make things more explicit and not having to guess how many params the func will take.

neilvyas:
I think adapt_call is fine for handling breaking legacy changes like this one. That said, I think the version of adapt_call that worked off of argspecs (and thus could force the user to specify an argspec to "adapt to") was nicer than the current approach, but I can understand portability issues.

PR #214, opened 2017-10-25
Setup `tox`

neilvyas:
For better local testing against multiple environments/python versions.

I heard something about micropython support so we could worm that in here, as well.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,10 @@
+[tox]
+skip_missing_interpreters=True
+envlist = py27,py33,py34,py35,py36,py37
+
+[testenv]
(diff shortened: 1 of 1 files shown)


mannygit:
@neilvyas this might be useful https://github.com/tox-dev/tox-travis 

It would allow for the travis config to be the single specification of environments but allow tox based testing locally.

dbrattli:
Closing this PR since it has become old and needs to be updated. We are e.g not using nose anymore. Please fix and reopen if you still want to setup tox with RxPY. The project is more active now and there should be a better chance of getting the PR reviewed.

PR #232, opened 2018-03-19
add blocking_subscribe()  operator

shyam-s00:
PR for adding blocking_subscribe() op


dbrattli:
This is very interesting. I will need to align it with aioreactive that have a similar method called 'run()'. So this will probably have to go into the vnext branch instead. Hopefully I can fix this over easter. Thanks!

PR #243, opened 2018-05-25
add LICENSE to distributions

bollwyvl:
Thanks for the work!

This adds the `LICENSE` to distributions, as per the [Apache policy](http://www.apache.org/legal/src-headers.html).

Background: I am looking to package this for conda via the conda-forge process. In the meantime, I'll include a copy of the license there!

Thanks again!

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include LICENSE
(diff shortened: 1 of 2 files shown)


bollwyvl:
Hooray!

Issue #169, opened 2017-03-15
observe_on(AsyncIOScheduler()) doesn't work for emissions from other threads


adamples:
Actually it can work quite randomly for more complicated programs. Basically when events are emitted using any scheduler that spawns threads (so `NewThreadScheduler`, `ThreadPoolScheduler`, `TimeoutScheduler`) the observer callbacks are not executed until some other event happens.

Example code (DebugTAP is a class that just prints received data):

    loop = asyncio.get_event_loop()

    # events will be emitted using TimeoutScheduler
    rx.Observable.interval(100) \
        .take(5) \
        .observe_on(AsyncIOScheduler(loop)) \
        .subscribe(DebugTAP())

    # nothing happens for 2 seconds
    loop.run_until_complete(asyncio.sleep(2))

    # here the numbers are printed, all at once, because loop was woken up
    loop.run_until_complete(asyncio.sleep(2))

I traced this problem down to the interaction between `ScheduledObserver` and `AsyncIOScheduler`. The issue is that `ScheduledObserver` schedules its `run` method on the target scheduler _but in the context of the source scheduler_, so possibly in a thread other than the one where the loop is running. `AsyncIOScheduler` in turn calls loop's `call_soon` and `call_later` methods, which are not thread-safe [1]. The result is that the loop is not aware of the new task that was just scheduled and waits for other events (like I/O or end of sleep as in my example).

The issue would be much less apparent in a program where the loop is constantly waken up, but then still unneccessary latency is introduced.

The simple solution would be to replace `call_soon` and `call_later` with `call_soon_threadsafe` and `call_later_threadsafe` respectively [2].

[1] https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.AbstractEventLoop
[2] https://docs.python.org/3/library/asyncio-dev.html#concurrency-and-multithreading

NavidHt:
If we enable asyncio debug mode it throws `RuntimeError` exception with a message complaining about **thread-safety**:
```
Exception in thread Thread-1:
Traceback (most recent call last):
  ...
  File ".../rx/concurrency/mainloopscheduler/asyncioscheduler.py", line 28, in schedule
    handle = self.loop.call_soon(interval)
  File "/usr/lib64/python3.6/asyncio/base_events.py", line 576, in call_soon
    self._check_thread()
  ...
RuntimeError: Non-thread-safe operation invoked on an event loop other than the current one
```
[...]

PR #249, opened 2018-06-29
Use threadsafe asyncio methods in AsyncIOScheduler

adamples:
This commit fixes https://github.com/ReactiveX/RxPY/issues/169

The problem was that schedule* methods can be called from other threads (e.g. created by other schedulers) and because of this thread-safe version of asyncio scheduling methods (call_soon_threadsafe and call_later_threadsafe) should be used.

### DIFF
--- a/rx/concurrency/mainloopscheduler/asyncioscheduler.py
+++ b/rx/concurrency/mainloopscheduler/asyncioscheduler.py
@@ -25,7 +25,7 @@ class AsyncIOScheduler(SchedulerBase):
 
         def interval():
             disposable.disposable = self.invoke_action(action, state)
-        handle = self.loop.call_soon(interval)
+        handle = self.loop.call_soon_threadsafe(interval)
 
         def dispose():
             handle.cancel()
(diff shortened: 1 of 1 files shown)


MainRo:
maybe using call_soon_threadsafe instead of call_soon should not be systematic: Using call_soon_threadsafe probably adds an overhead. So we could add a parameter to the AsyncIOScheduler constructor to choose if the scheduler must be threadsafe or not ?

something like:

__init__(self, loop=None, thread_safe=False)

adamples:
Yes, I think configuration option is a good idea.

But this is fix is even more complicated than I anticipated.

For some reason there is no `call_later_threadsafe`, only `call_soon_threadsafe`. One solution might be something like `loop.call_soon_threadsafe(loop.call_later, ...)` but there is a problem that cancellation is then much more difficult.

Additionally `handle.cancel()` should be scheduled with `call_soon_threadsafe` but that would mean we can't be sure that task is cancelled after `dispose` returns (it may be some time later).

Unfortunately I don't have enough time now to implement the proper fix, so I'm closing this PR.

PR #266, opened 2018-11-29
implemented threadsafe schedule methods

### LINKED ISSUE #169: observe_on(AsyncIOScheduler()) doesn't work for emissions from other threads

MainRo:
This commit fixes #169.

The immediate scheduling method is based on call_soon_threadsafe to
schedule the action on the event loop. The dispose function uses
a future from the concurrent package so that it can block until the
action is canceled.

The relative scheduling method is executed in two steps: The first one
allows to move the execution context to the event loop. The second step
schedules the action at the appropriate time.

### DIFF
--- a/rx/concurrency/mainloopscheduler/asyncioscheduler.py
+++ b/rx/concurrency/mainloopscheduler/asyncioscheduler.py
@@ -1,6 +1,7 @@
 import logging
 import asyncio
 
+from concurrent.futures import Future
 from rx.core import Disposable
 from rx.disposables import SingleAssignmentDisposable, CompositeDisposable
 from rx.concurrency.schedulerbase import SchedulerBase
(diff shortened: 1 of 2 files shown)


dbrattli:
Thanks!

PR #274, opened 2019-01-09
partial tests update to pipe style

jcafhe:
Tests update for operators : 

- all
- average
- distinct_until_changed
- do_action
- flat_map
- last
- materialize

There are still some issues : 
- in all() tests: missing _some()_ operator (not in the test)
- in flat_map() tests: missing _interval()_, I have assumed it will be in the rx.init API instead of rx.operators
- in flat_mapi() tests: the old test provides a second kwarg  'result_mapper_indexed' in test_flat_map_iterable_complete_result_mapper
- in materialize() tests: missing _dematerialize()_ operator  

I have used '_' as an alias to rx.operators (saw this in the catch_exception test), but I've switched to 'ops' to avoid possible conflict. I don't know if a convention is needed ?

### DIFF
--- a/tests/test_observable/test_all.py
+++ b/tests/test_observable/test_all.py
@@ -1,5 +1,6 @@
 import unittest
 
+from rx import operators as _
 from rx.testing import TestScheduler, ReactiveTest
 
 on_next = ReactiveTest.on_next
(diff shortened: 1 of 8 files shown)


dbrattli:
As with RxJS 6, all result mappers will eventually be removed. This is not a problem since you can always use a map operator on the result if needed. Thus all tests using result mappers should just be removed.

dbrattli:
I've also been thinking about using `ops` instead of `_`. It is probably better self describing than `_`. We just need to see what looks best I guess so we can agree on a convention. Not a problem as it should be easy to do a global search replace later.

PR #275, opened 2019-01-10
update tests: amb, of, case, for_in, some, generate, switch_latest, timestamp & time_interval

jcafhe:
Update tests to pipe style for operators:
- amb
- of
- case
- for_in
- some
- generate 
- switch_latest
- timestamp & time_interval


dbrattli:
Thanks a lot!

PR #276, opened 2019-01-10
Fix forin and some operators

jcafhe:
Fixing the following operators: 
-  __for_in__: wrong import
- __some__: using pipe instead of chain


dbrattli:
Thanks a lot!

PR #277, opened 2019-01-11
update tests: min, min_by, max, max_by, zip, delay_with_mapper, contains, sum, to_async, skip, skip_last, default_is_empty, pairwise, using, window

jcafhe:
Tests update to pipe style for operators:

- min, min_by, max, max_by,
- zip
- delay_with_mapper (__IMPORTANT__: I've assumed delay_with_mapper was delay_with_selector before, so I've pushed 2 commits, one for pipe a74584d42de6002ea29255b7065f8f54b5a8b647,  second for renaming the file 5865c579963e69f75164248304f28c39b9cfa135)
- contains
- sum
- to_async
- skip, skip_last
- default_is_empty
- pairwise
- using
- window

I am actually a bit worry about how the list of commits per PR is growing even if previous PRs were accepted. It seems to be related to the fact that I'm keeping my branch 'test_to_pipestyle' alive to make PRs. Ocasionnaly, I use a merge to keep it synced with feature/rxpy-3.0.  Not sure if it's the right way to do things. Should I rebase instead of merge after a PR is accepted?


dbrattli:
Thanks, and great work! Don't worry about the PRs. We just need to make this as easy as possible. They are also going into a feature branch which will be  merged into master at one point, and then we can rewrite things the way we want.

PR #279, opened 2019-01-13
tests update: take, take_last, concat, defer, from_, timer, interval, take_with_time, combine_latest, throttle_first

jcafhe:
Tests update to pipe style for operators:
- test_take          
- take_last
- concat
- defer
- from_
- timer
- interval
- take_with_time
- combine_latest
- throttle_first


dbrattli:
Awesome!

PR #281, opened 2019-01-13
Fix combine_latest:  mapper should be provide as a keyword argument

jcafhe:
When calling rx.combine_latest(*args, mapper), mapper must be a keyword argument because _mapper_ argument is defined after the positionnal arguments *args.


dbrattli:
Nice catch!

PR #286, opened 2019-01-21
remove result_mapper for zip_with_iterable operator

jcafhe:
Remove result_mapper argument for zip operator. 

The resulting elements are now zipped in tuple.

### DIFF
--- a/rx/core/operators/zip.py
+++ b/rx/core/operators/zip.py
@@ -26,26 +26,22 @@ def _zip(*args: Observable) -> Callable[[Observable], Observable]:
         return rx.zip(*sources)
     return zip
 
-def _zip_with_iterable(second, result_mapper):
+def _zip_with_iterable(second):
     def zip_with_iterable(source: Observable) -> Observable:
         """Merges the specified observable sequence and list into one
-        observable sequence by using the mapper function whenever all of
+        observable sequence by using a tuple aggregation whenever all of
         the observable sequences have produced an element at a
         corresponding index.
 
-        The result mapper must be a function to invoke for each series of
-        elements at corresponding indexes in the sources.
-
         Example
-            >>> res = zip(xs, [1,2,3], result_mapper=fn)
+            >>> res = zip(xs, [1,2,3])
 
         Args:
             source -- Source observable to zip.
 
         Returns:
             An observable sequence containing the result of
-            combining elements of the sources using the specified result
-        mapper function.
+            combining elements of the sources as tuple.
         """
 
         first = source
@@ -61,7 +57,8 @@ def _zip_with_iterable(second, result_mapper):
                     right = second[index]
                     index += 1
                     try:
-                        result = result_mapper(left, right)
+                        result = (left, right)
+
                     except Exception as ex:
                         observer.on_error(ex)
                         return
--- a/rx/core/operators/zip.py
+++ b/rx/core/operators/zip.py
@@ -56,12 +56,7 @@ def _zip_with_iterable(second):
                 if index < length:
                     right = second[index]
                     index += 1
-                    try:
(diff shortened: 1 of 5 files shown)


dbrattli:
I really like this, returning a tuple instead of using the result_mapper, and then composing with map afterwards. It is btw exactly how I do it in the  implementation of [zipSeq](https://github.com/dbrattli/Reaction/blob/master/AsyncRx/src/Combine.fs#L243) in Reaction.AsyncRx (F#). We should btw do the same with `combine_latest` and `with_latest_from` (imo).

rx/core/operators/zip.py
```diff
@@ -61,7 +57,8 @@ def on_next(left):
                     right = second[index]
                     index += 1
                     try:
-                        result = result_mapper(left, right)
+                        result = (left, right)
+
                     except Exception as ex:
```
dbrattli: Just remove the try/except as well. Not needed anymore.

rx/operators/__init__.py
```diff
@@ -2550,32 +2550,25 @@ def zip(*args: Observable) -> Callable[[Observable], Observable]:
     return _zip(*args)
 
 
-def zip_with_iterable(second, result_mapper):
+def zip_with_iterable(second):
     """Merges the specified observable sequence and list into one
-    observable sequence by using the mapper function whenever all of
+    observable sequence by using a tupple aggregation whenever all of
```
dbrattli: by creating a tuple whenever all of ...

rx/operators/__init__.py
```diff
@@ -2550,32 +2550,25 @@ def zip(*args: Observable) -> Callable[[Observable], Observable]:
...
-        result_mapper: Mapper function that produces an element
-            whenever all of the observable sequences have produced an
-            element at a corresponding index
 
     Returns:
         An operator function that takes and observable source and
         returns an observable sequence containing the result of
-        combining elements of the sources using the specified result
-        mapper function.
+        combining elements of the sources as tuple.
```
dbrattli: as a tuple

tests/test_observable/test_zip.py
```diff
@@ -365,9 +365,10 @@ def test_zip_with_iterable_never_empty(self):
         n2 = []
 
         def create():
-            def mapper(x, y):
-                return x + y
-            return n1.pipe(ops.zip_with_iterable(n2, result_mapper=mapper))
+            return n1.pipe(
+                    ops.zip_with_iterable(n2),
+                    ops.map(sum),
```
dbrattli: Indent not correctly (should be 4 spaces)

tests/test_observable/test_zip.py
```diff
@@ -380,9 +381,10 @@ def test_zip_with_iterable_empty_empty(self):
...
+                    )
```
dbrattli: Indent not correctly (should be 4 spaces)

rx/core/operators/zip.py
jcafhe: Nice catch, we should do the same for the zip operator.

tests/test_observable/test_zip.py
jcafhe: this should be fixed now.

jcafhe:
I agree, returning tuples from `combine_latest`& `with_latest_from` would be great. So you would remove the mapper function from these operators?

The only drawback I see, if `combine_latest` returns a tuple and is piped with `map`, is that you can't provide anymore a function with positional arguments. You need to unpack the tuple in the function body.

dbrattli:
Yes, I see the problem. Perhaps you would like to implement a new operator called `starmap`?  Inspired by https://docs.python.org/2/library/itertools.html#itertools.starmap

PR #288, opened 2019-01-22
Remove result_mapper for join and group_join operators

jcafhe:
Removes *result_mapper* argument for `join` & `group_join` operators. As `zip`, elements are zipped in a tuple.

Regarding `group_join`, the try/except is still there because of the call to the function *add_ref*.

Tests relying on an exception raised in *result_mapper* are removed.


dbrattli:
Great job, thanks!

PR #289, opened 2019-01-22
remove (result) mapper for combine_latest operator

jcafhe:
Remove *mapper* argument for `combine_latest` operator. Elements are now zipped in a tuple.


dbrattli:
Great job, thanks!

PR #292, opened 2019-01-23
remove (result) mapper for with_latest_from operator

jcafhe:
Remove mapper argument for combine_latest operator. Elements are now zipped in a tuple.

The last one ! :tada:


dbrattli:
Nice work!

PR #294, opened 2019-01-23
feature: implement starmap operator and tests

jcafhe:
Adding `starmap` operator as discussed.

- I didn't create a implementation file in rx/core/operators since it is simple.
- Tests are inspired by `map` tests + some other cases.
- Do you think it's ok to name it `starmap` instead of `star_map` ? I prefer `starmap` but it may not be very pythonic.
- I'm not very happy with the docstring, my english is not very good.


dbrattli:
I was hoping it to be a simple implementation, so nice work! I will look at the docstring and make fixes if I see anything.

PR #298, opened 2019-01-28
Restore rx.streams in setup.py

kobusvdm:


kobusvdm:
Apologies. I see the missing rx.streams is due to the start of the Rx 3.0 iteration. I'll close this

PR #303, opened 2019-02-10
remove iterable argument for concat operator

jcafhe:
This PR is for removing the first argument of `concat` operator which accepts an Iterable[Observable].

A new operator `rx.concat_with_iterable` has been added for operators which need to concat an iterable of observables:
- `repeat`
- `for_in`
- `while_do`


dbrattli:
This is awesome! Thanks, and keep up the great work!

jcafhe:
Thanks!

PR #306, opened 2019-02-13
Schedulers, CI

erikkemperman:
This PR has some small fixes in the mainloop schedulers, and expands the Travis setup to test against several versions of Python on Linux, as well as 3.7-dev on MacOS and Windows (the latter is slightly experimental, but it is working well with this particular configuration and is expected to gain official support in the future.)

For the base-line, Linux Python 3.6, some additional dependencies are installed (eventlet gevent tornado twisted, pycairo and pygobject) so that the coverage increases a bit for the mainloop schedulers. I tried to add either pyside2 or pyqt5 in the same way but that went horribly wrong (core dumps). I haven't figured out why yet, it works fine on my wokstation.

I noticed some mainloop schedulers are not tested at all at the moment (wx, pygame) but didn't have time to write anything for them. I actually would like to propose some refactoring of the scheduler tests, there seems opportunity for some re-use, but I will keep that separate of this PR.

Anyway, I hope this is going in the right direction -- but if you would like me to amend anything please let me know!

### DIFF
--- a/tests/test_concurrency/test_mainloopscheduler/test_tkinterscheduler.py
+++ b/tests/test_concurrency/test_mainloopscheduler/test_tkinterscheduler.py
@@ -5,9 +5,9 @@ import pytest
 from rx.concurrency.mainloopscheduler import TkinterScheduler
 
 tkinter = pytest.importorskip("tkinter")
-
+import tkinter
 try:
-    root = tkinter.Tk()
+    root = tkinter.Tcl()
     display = True
 except Exception:
     display = False
--- a/tests/test_concurrency/test_mainloopscheduler/test_tkinterscheduler.py
+++ b/tests/test_concurrency/test_mainloopscheduler/test_tkinterscheduler.py
@@ -5,7 +5,7 @@ import pytest
 from rx.concurrency.mainloopscheduler import TkinterScheduler
 
 tkinter = pytest.importorskip("tkinter")
-import tkinter
+
 try:
     root = tkinter.Tcl()
     display = True
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,42 @@
-language: python
-python:
-  - "3.6"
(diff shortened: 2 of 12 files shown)


dbrattli:
Great work. Really like this PR!

As for refactoring the tests, I don't want any "clever" code or abstractions as it might make the tests harder to read. For tests I don't mind "repeating myself" since it should be easy to add or remove stuff for one scheduler without breaking every other scheduler.

erikkemperman:
Cool, glad you like my first PR. Point taken about refactoring the tests.

tests/test_concurrency/test_mainloopscheduler/test_tkinterscheduler.py
```diff
@@ -5,9 +5,9 @@
 from rx.concurrency.mainloopscheduler import TkinterScheduler
 
 tkinter = pytest.importorskip("tkinter")
-
+import tkinter
```
erikkemperman: Ugh, sorry, this import should not have been added. I just pushed an extra commit to remove it.

erikkemperman:
By the way, it looks like coveralls isn't reporting back to github -- even though the travis log says it's pushed the results. Let me see if I can figure out why that is.

https://travis-ci.org/ReactiveX/RxPY/jobs/492870377#L1310

erikkemperman:
Well, it was warning about missing pyyaml before. I got rid of that warning now, but it doesn't seem to do the trick?

erikkemperman:
Hmm, somehow it seems to go to the wrong place. The travis logs say it was pushed here: https://coveralls.io/jobs/45160328
But that falls under dbrattli/rxpy instead of reactivex/rxpy. Recent builds have proper coveralls reporting, and I don't think anything I did could have impacted this. Unfortunately I have to sign off for today, and will be outdoors all day tomorrow...

erikkemperman:
Looks like coveralls reported back after all. Thanks a lot for review and merge!

Issue #270, opened 2019-01-05
RxPY 3.0: Documentation


MainRo:
With v3, an official documentation should be provided. This issue can be used to follow the progress on this topic. The following things must be done

- [X] create outline
- [X] migrate part of readme in doc
- [X] reference doc from readme
- [X] add a migration section
- [X] document APIs : autodoc for all operators and classes
- [X] choose a theme : Guzzle. 
- [x] publish the doc somewhere:  https://rxpy.readthedocs.io
- [x] add marble diagrams for operators
- [X] add additional readings section
- [X] license section
- [X] contributing section
- [X] Fix docstring errors in __init__.py
- [x] spellcheck
- [ ] final pass on each section

dbrattli:
There are some theme samples are http://www.writethedocs.org/guide/tools/sphinx-themes/ . Read-the-docs is nice. Also the Guzzle theme looks quite nice and clean.

As for hosting we should be able to get permission to deploy at e.g. http://reactivex.io/rxpy

MainRo:
I have several commits ready. Should I push them on feature/rxpy-3.0 or wait until the merge on master is done ?

MainRo:
Concerning the marble diagrams, initially I wanted to reuse the ones from reactivex.io. However they are not all consistent (there are several graphic styles, depending on the operators). Moreover linking to picture does not allow to see the marble diagram when reading the source code. 

So I wonder if the following could be possible: Create a sphinx addon that would use a tool to convert text based marble diagram to pictures. This would have several advantages compared to static pictures that are hard to update:

- The diagrams are readable in the code documentation.
- They can be easily updated.
- diagrams are automatically available when adding new operators
- diagrams style is consistent.

The main issue here is to get this marble renderer. Surprisingly I did not find such a tool. If anybody is aware of an existing tool, I am interested! Otherwise maybe it could be relatively easy to implement one based on matplotlib. I will try some mocks if we want to document marble diagrams this way.

dbrattli:
We might want to merge with master first. It's now being added new features on the feature branch, so it's time to merge. We can still have a period of rapid development before locking down and starting doing PRs for the master branch.

[...]

PR #308, opened 2019-02-18
Feature/marble

MainRo:
This PR contains three examples of marbles diagrams that are generated with dooble and its sphinx extension: map, catch, window. The three examples cover all features supported by dooble.

This PR is for continuation of #270


dbrattli:
I think this looks really nice. Great work!

PR #317, opened 2019-02-27
Fix timeflies examples for GTK, QT, WX

jcafhe:
This PR fixes timeflies examples for GTK, Qt & Wx. 

## Qt:
- fix: imports, pipe, time in seconds.

## GTK:
- fix: imports, pipe, time in seconds.
- fix: main window is constantly auto resized. The width/height of the main window keep growing when the mouse cursor gets close to the right or bottom borders. The labels which are drawn outside of the main window force an auto resize of the container. Quick fix: wrap the container in a `Gtk.ScrolledWindow`.
- Deprecation: specify *label* as a keyword argument in `Gtk.Label` constructor.

```python
/home/jehf/DEV/pyprojects/RxPY/examples/timeflies/timeflies_gtk.py:38: PyGTKDeprecationWarning: Using positional arguments with the GObject constructor has been deprecated. Please specify keyword(s) for "label" or use a class specific constructor. See: https://wiki.gnome.org/PyGObject/InitializerDeprecations
  labels = [Gtk.Label(char) for char in text]
```

## Wx:
- fix: imports, pipe, time in seconds.
- Deprecation: change MoveXY() to Move()
```python
/home/jehf/DEV/pyprojects/RxPY/examples/timeflies/timeflies_wx.py:38: wxPyDeprecationWarning: Call to deprecated item. Use Move instead.
  label.MoveXY(x + i*12 + 15, y)
 ```
## Wx scheduler:
- fix `msecs` variable referenced before assignment.

### DIFF
--- a/examples/timeflies/timeflies_gtk.py
+++ b/examples/timeflies/timeflies_gtk.py
@@ -1,6 +1,6 @@
+from rx import operators as ops
 from rx.subjects import Subject
-from rx.concurrency import GtkScheduler
-import sys
+from rx.concurrency.mainloopscheduler import GtkScheduler
 
 import gi
 gi.require_version('Gtk', '3.0')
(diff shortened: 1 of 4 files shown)


dbrattli:
Nice, as a bonus you could also consider making a PR for rewriting the `for` loop into a `flat_map` to make it even more reactive. Check out `timeflies_tkinter.py`.

jcafhe:
Yes, this seems much cleaner.

Issue #241, opened 2018-04-29
compatibility for upcoming PySide2-5.11 preview


stonebig:
After 2 years of brewing, PySide2-5.11 will reach preview status in may 2018, with:
- full support per the Qt Company,
- full API support at release time of Qt versions.

Would it be possible to have PySide2 compatibility for that may 2018 milestone ?
(QtPy, vendored or not,  being a possible intermediate layer)

see http://blog.qt.io/blog/category/qt-for-python/

for PySide2 bug tracker, see https://wiki.qt.io/Qt_for_Python

stonebig:
my old example doesn't currently work:
````
# from https://github.com/ReactiveX/RxPY/blob/master/examples/timeflie
from rx.subjects import Subject
from rx.concurrency import QtScheduler
import sys

try:
    from PyQt4 import QtCore
    from PyQt4.QtGui import QWidget, QLabel
    from PyQt4.QtGui import QApplication
except ImportError:
    try:
        from PyQt5 import QtCore
        from PyQt5.QtWidgets import QApplication, QWidget, QLabel
    except ImportError:
        from PySide import QtCore
        from PySide.QtGui import QWidget, QLabel
        from PySide.QtGui import QApplication


class Window(QWidget):

    def __init__(self):
        super(QWidget, self).__init__()
        self.setWindowTitle("Rx for Python rocks")
        self.resize(600, 600)
        self.setMouseTracking(True)

        # This Subject is used to transmit mouse moves to labels
        self.mousemove = Subject()

    def mouseMoveEvent(self, event):
        self.mousemove.on_next((event.x(), event.y()))


def main():
    app = QApplication(sys.argv)
    scheduler = QtScheduler(QtCore)

    window = Window()
    window.show()

    text = 'TIME FLIES LIKE AN ARROW'
    labels = [QLabel(char, window) for char in text]

    def handle_label(i, label):

        def on_next(pos):
            x, y = pos
            label.move(x + i*12 + 15, y)
            label.show()

        window.mousemove.delay(i*100, scheduler=scheduler).subscribe(on_next)

    for i, label in enumerate(labels):
        handle_label(i, label)

    sys.exit(app.exec_())

if __name__ == '__main__':
    main()
````

leinardi:
Any news about supporting PySide2?

baudren:
As far as I can tell, it's working. I fixed the imports and, for some reason, the init call to the parent class, and I get the text following my arrow on screen.

[...]

PR #318, opened 2019-02-27
Refactor timeflies Qt, WX, GTK 

jcafhe:
Rewritting timeflies examples in a more 'reactive' way as suggested in PR #317.

 Regarding timeflies for Qt, `Pyside2` has been added. PyQt5 & Pyside2 are confirmed to work. This is related to #241.


dbrattli:
Fantastic!

PR #319, opened 2019-02-28
QtScheduler

erikkemperman:
I made the import precedence for QtScheduler tests consistent with the example, and I fixed an error as well (schedule_absolute was throwing overflow errors). Added a test case for that.

Tried to get PyQt working in Travis as well, but I am still getting core dumps for some reason... Locally all works very well. But I did wind up making the Travis config just a little bit cleaner, the builds on Mac and Windows now use regular 3.7 instead of odd 3.7-dev versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,14 @@ matrix:
       python: 3.6
       # Coverage for the baseline 3.6 build: include some optional packages
       before_script:
-        - pip install eventlet gevent pyyaml tornado twisted
+        - pip3 install eventlet gevent pyyaml tornado twisted
         # pycairo / pygobject need native libraries
         - sudo apt-get install -y libgirepository1.0-dev gir1.2-gtk-3.0
-        - pip install pycairo pygobject
+        - pip3 install pycairo pygobject
+      script:
+        - coverage run --source=rx setup.py test
+      after_success:
+        - coveralls
 
     - os: linux
       dist: xenial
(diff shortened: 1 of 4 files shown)


jcafhe:
Following your  [comment](https://github.com/ReactiveX/RxPY/commit/1ac4c86c4b3bded55f791752e88aceab8ec4d8f0#r32515003), maybe we should simply remove Pyside(1) support in the timeflies example since RxPY v3 would require python 3.6 (at least?). 

Also, I see that you're working actively on schedulers. I was planning to fix a leak with the `QtScheduler` (related to `self._timers`) and do some refactoring too. This will be quick but is it ok for you, as it might interfere with your work?

erikkemperman:
I removed pyside1 from the example in this PR.

My efforts on the schedulers are work in progress, I think I am finding some little improvements here and there. Mostly I’m trying to prepare the ground, as it were, to submit a ProcessPoolScheduler.

But anyway, please just go ahead and submit what you have ready — I’ll figure out how to rebase my stuff.

jcafhe:
A ProcessPoolScheduler could be awesome :+1:  However, I guess implementing it would not be so easy, especially dealing with limitations of pickle serialization .

By the way, do you know if pyqt4 works with python >= 3.6? I can't find any clear answer about that.

erikkemperman:
Yes, it will be tricky, and might not succeed... I am looking at dill for serialization, if you were wondering.

The core dumps happen during test, install is fine. Thanks, I will try if relaxing timings might help.

Not sure about qt4 compatibility, I’ll let you know if I find out.

PR #321, opened 2019-03-02
Fix: do_action operator does not forward downstream scheduler + fix do operator

jcafhe:
Small Fixes:
- forward scheduler from `subscribe` method to the source observable
- re-implement `do` operator


dbrattli:
Nice, see that we need to fix the do_after_next also, but tempted to remove a lot of these obscure and less used operators.

jcafhe:
Yes, was tempted too :D 
In any case, these operators are not exposed in `operators.__init__`

PR #324, opened 2019-03-12
Scheduler tweaks

erikkemperman:
I wanted to submit this now, before things get even further out of hand...

Looking at the priority queue and the "basic" schedulers (the ones used by operators) I believe I found a few improvements. I tried to keep each commit self-explanatory, and went to the trouble of running all of them through Travis incrementally. You can see everything passes, [here](https://github.com/erikkemperman/RxPY/commits/scheduler-tweaks).

But in total this is perhaps a little too big for a single PR. However, since each commit builds on the previous, I did not think it would be feasible to submit multiple PRs. Sorry about that. If it helps, reviewing the commits one at a time might be easier than reviewing the whole diff at once.

I've added a crude benchmark at the end -- basically timing the unit tests minus test_concurrency several times -- to demonstrate that these changes (particularly to `CurrentThreadScheduler` and `EventLoopScheduler`) offer a significant perfomance boost.

This is mostly due to preventing a bunch of threads where they aren't strictly needed -- threads are cheap on Linux, not so on Mac or Windows. Compared to the master branch:

~6% faster on [Linux](https://travis-ci.org/erikkemperman/RxPY/jobs/504088612#L1087)
~21% faster on [MacOS](https://travis-ci.org/erikkemperman/RxPY/jobs/504088614#L985)
~43% faster on [Windows](https://travis-ci.org/erikkemperman/RxPY/jobs/504088615#L949)

Additionally, I polished the mainloop schedulers a bit in terms of typing and docstrings.

I would like to propose additional tweaks to the `TimeoutScheduler`, which could also prevent a lot of spurious threads. Or alternatively make a new scheduler, something like `TimerPoolScheduler`, to be used internally by the operators. But that will have to wait and I didn't want to keep postponing submitting what I have at this point.

### DIFF
--- a/rx/concurrency/mainloopscheduler/qtscheduler.py
+++ b/rx/concurrency/mainloopscheduler/qtscheduler.py
@@ -105,7 +105,7 @@ class QtScheduler(SchedulerBase):
             The disposable object used to cancel the scheduled action
             (best effort).
         """
-        msecs = int(self.to_seconds(period)) * 1000
+        msecs = int(self.to_seconds(period) * 1000.0)
         sad = SingleAssignmentDisposable()
 
         periodic_state = state
--- a/rx/concurrency/mainloopscheduler/qtscheduler.py
+++ b/rx/concurrency/mainloopscheduler/qtscheduler.py
@@ -1,9 +1,12 @@
 import logging
 
-from rx.disposable import Disposable
+from typing import Optional
+
 from rx.core import typing
-from rx.disposable import SingleAssignmentDisposable, CompositeDisposable
-from rx.concurrency.schedulerbase import SchedulerBase
+from rx.disposable import CompositeDisposable, Disposable, SingleAssignmentDisposable
+
+from ..schedulerbase import SchedulerBase
+
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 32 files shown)


rx/concurrency/mainloopscheduler/qtscheduler.py
```diff
@@ -51,39 +69,43 @@ def dispose():
...
+            state: [Optional] state to be given to the action function.
 
         Returns:
-            The disposable object used to cancel the scheduled
-            recurring action (best effort).
+            The disposable object used to cancel the scheduled action
+            (best effort).
         """
-        msecs = int(self.to_seconds(period)*1000.0)
+        msecs = int(self.to_seconds(period)) * 1000
```
jcafhe: Should it be `msecs = int(self.to_seconds(period) * 1000)` instead? 
E.g. I feel that 1.6s will be rounded to 1s resulting in 1000ms instead of 1600ms.

erikkemperman: Ah, thanks! I accidentally deleted a parenthesis here and restored it in the wrong place... Will amend immediately.

erikkemperman:
Sorry for the inconvenience, but I just submitted a slightly more focussed PR [here](https://github.com/ReactiveX/RxPY/pull/327). I will close this one, accordingly.

Issue #316, opened 2019-02-23
Coveralls reports


erikkemperman:
It seems coveralls as run in Travis (sometimes) pushes its results to the wrong place, namely dbrattli/rxpy instead of reactivex/rxpy, and consequently they are not reported back to github correctly.

I’m not totally sure, but I think it might be confused by the repo tokens in .coveralls.yml — these are identical for those two repositories. I would suggest generating a new token for the main repository, if I’m right about the cause of this, it may resolve it.

PR #327, opened 2019-03-15
PriorityQueue and related Schedulers

erikkemperman:
This supercedes https://github.com/ReactiveX/RxPY/pull/324 -- I've re-considered and decided to put in a little extra effort to submit a somewhat more topical PR.

In particular, I've left out some cosmetic changes which were not directly linked to the PriorityQueue or its related schedulers. They are entirely trivial changes, but polluted the already too large diff because they touch many files.

If we can reach agreement on this PR (I'd be happy to amend in response to suggestions or complaints, of course) I will submit the remaining polish commits at some later time.

As before, I've run every commit separately through Travis, you can see that all of them pass [here](https://github.com/erikkemperman/RxPY/commits/priorityqueue). I believe that the least annoying way to review this is one commit at a time -- nevertheless, I apologize that the PR is still pretty unwieldy...

I also figured I ought to be more verbose in motivating these changes... All links in the following refer to the current master branch, at [b3334408](https://github.com/ReactiveX/RxPY/tree/b333440871ad5e74dbdf5365fe99643f22f92c25):

- I noticed that the PriorityScheduler has some minor, superficial, quirks:
  - It declares a [`capacity`](https://github.com/ReactiveX/RxPY/blob/b333440871ad5e74dbdf5365fe99643f22f92c25/rx/internal/priorityqueue.py#L11) constructor arg, but it is not enforced. I prefer to err on the side of simplicity, so for the moment I propose removing the argument.
  - It declares a [`remove_at`](https://github.com/ReactiveX/RxPY/blob/b333440871ad5e74dbdf5365fe99643f22f92c25/rx/internal/priorityqueue.py#L29) method, but addressing the queue by index isn't a well-defined operation (note that while sorting the list maintains the heap invariant, but conversely a valid heap does not necessarily correspond to a sorted list).
  - The [`peek`](https://github.com/ReactiveX/RxPY/blob/b333440871ad5e74dbdf5365fe99643f22f92c25/rx/internal/priorityqueue.py#L27) method raises a custom InvalidOperationException when the queue is empty, but other methods don't and simply raise IndexError in the same circumstance.
  - It ought to be Generic, in the type of items in the queue.

- The methods of PriorityQueue are synchronized "coarsely", but I notice there is room to be a little more precise with locking the queues, if we define the queue itself to be unsynchronized and shift the burden to the classes that use it:
  - The CurrentThreadScheduler, it seems to me, should not need any explicit locks at all -- its own semantics ought to guarantee safety. But currently it is imposing a [second layer](https://github.com/ReactiveX/RxPY/blob/b333440871ad5e74dbdf5365fe99643f22f92c25/rx/concurrency/currentthreadscheduler.py#L46) of synchronization, on top of the queue's own lock.
  - Also, the CurrentThreadScheduler uses a [dict](https://github.com/ReactiveX/RxPY/blob/b333440871ad5e74dbdf5365fe99643f22f92c25/rx/concurrency/currentthreadscheduler.py#L45) to make sure that each invoking thread is mapped to its own queue. But that is just what `threading.local` does, no need to implement it manually (and leaking keys in the dict, as threads disappear).
  - Refactoring to address the above, we find that there is no longer a need for re-entrant locks and we can use the cheaper regular lock.

- Similar to the preceding arguments, more or less straightforward changes can be made to the other usages of the PriorityQueue, namely the EventLoopScheduler, VirtualTimeScheduler and PyGameScheduler.
  - The EventLoopScheduler, moreover, currently uses a separate [`Timer`](https://github.com/ReactiveX/RxPY/blob/b333440871ad5e74dbdf5365fe99643f22f92c25/rx/concurrency/eventloopscheduler.py#L145) which is unnecessary. Since threads are costly on some platforms, I propose avoiding them when it is easy to do so.
   - When using regular (non re-rentrant) locks, we have to be careful to call the `invoke` methods of scheduled items outside of the synchronized / critical sections -- otherwise recursive scheduling would go amiss. Fortunately the changes needed to achieve this for the affected schedulers aren't too complicated.

By the way, I'm not sure I understand the PyGameScheduler properly: I've made sure that I maintain the semantics that were before, but it seems odd to me that this scheduler doesn't actually rely on any pygame primitives?

Finally, I did include some polishing, but only on the files that were affected already.

Also, to get an idea of the impact on coverage, it might be nice to quickly try what I suggested in issue [#316](https://github.com/ReactiveX/RxPY/issues/316) -- it should take a couple of minutes, and there's no harm in trying since coverage reporting is currently not working?
 
Again, I apologize profusely for the inconvenience of such a large PR -- I hope that you agree there might be some benefit to these changes.

### DIFF
--- a/rx/concurrency/virtualtimescheduler.py
+++ b/rx/concurrency/virtualtimescheduler.py
@@ -33,6 +33,12 @@ class VirtualTimeScheduler(SchedulerBase):
 
         super().__init__()
 
+    def _get_clock(self):
+        with self._lock:
+            return self._clock
+
+    clock = property(fget=_get_clock)
+
     @property
     def now(self) -> datetime:
         """Gets the schedulers absolute time clock value as datetime offset."""
@@ -97,7 +103,8 @@ class VirtualTimeScheduler(SchedulerBase):
                 return
             self._is_enabled = True
 
-        spinning = 0
+        spinning: int = 0
+
         while True:
             with self._lock:
                 if not self._is_enabled or not self._queue:
--- a/rx/concurrency/virtualtimescheduler.py
+++ b/rx/concurrency/virtualtimescheduler.py
@@ -1,6 +1,7 @@
 import logging
 from datetime import datetime
-from typing import Optional, Any
+import threading
+from typing import Any, Optional
 
 from rx.internal import PriorityQueue, ArgumentOutOfRangeException
 from rx.core import typing
@@ -11,6 +12,7 @@ from .scheduleditem import ScheduledItem
 log = logging.getLogger("Rx")
 
 MAX_SPINNING = 100
+
 
 class VirtualTimeScheduler(SchedulerBase):
     """Virtual Scheduler. This scheduler should work with either
@@ -28,6 +30,7 @@ class VirtualTimeScheduler(SchedulerBase):
         self.clock = initial_clock
 
         self.is_enabled = False
+        self.lock = threading.RLock()
         self.queue: PriorityQueue[ScheduledItem[typing.TState]] = PriorityQueue()
 
         super().__init__()
@@ -65,7 +68,8 @@ class VirtualTimeScheduler(SchedulerBase):
         """Schedules an action to be executed at duetime."""
 
         si: ScheduledItem[typing.TState] = ScheduledItem(self, state, action, duetime)
-        self.queue.enqueue(si)
+        with self.lock:
+            self.queue.enqueue(si)
         return si.disposable
 
     def start(self) -> None:
@@ -107,34 +111,34 @@ class VirtualTimeScheduler(SchedulerBase):
         Args:
             time: Absolute time to advance the schedulers clock to.
         """
+        with self.lock:
+            if self.clock > time:
+                raise ArgumentOutOfRangeException()
 
-        if self.clock > time:
-            raise ArgumentOutOfRangeException()
+            if self.clock == time:
+                return
 
-        if self.clock == time:
-            return
+            if self.is_enabled:
+                return
 
-        if self.is_enabled:
-            return
+            self.is_enabled = True
 
-        self.is_enabled = True
+            while self.is_enabled:
+                item = self.get_next()
+                if not item:
+                    break
 
-        while self.is_enabled:
-            item = self.get_next()
-            if not item:
-                break
+                if item.duetime > time:
+                    self.queue.enqueue(item)
+                    break
 
-            if item.duetime > time:
-                self.queue.enqueue(item)
-                break
+                if item.duetime > self.clock:
+                    self.clock = item.duetime
 
-            if item.duetime > self.clock:
-                self.clock = item.duetime
+                item.invoke()
 
-            item.invoke()
-
-        self.is_enabled = False
-        self.clock = time
+            self.is_enabled = False
+            self.clock = time
 
     def advance_by(self, time: float) -> None:
         """Advances the schedulers clock by the specified relative time,
@@ -168,10 +172,11 @@ class VirtualTimeScheduler(SchedulerBase):
     def get_next(self) -> Optional[ScheduledItem]:
         """Returns the next scheduled item to be executed."""
 
-        while self.queue:
-            item: ScheduledItem[typing.TState] = self.queue.dequeue()
-            if not item.is_cancelled():
-                return item
+        with self.lock:
+            while self.queue:
+                item: ScheduledItem[typing.TState] = self.queue.dequeue()
+                if not item.is_cancelled():
+                    return item
 
         return None
 
--- a/rx/concurrency/virtualtimescheduler.py
+++ b/rx/concurrency/virtualtimescheduler.py
@@ -28,7 +28,7 @@ class VirtualTimeScheduler(SchedulerBase):
         self.clock = initial_clock
 
         self.is_enabled = False
-        self.queue = PriorityQueue()
+        self.queue: PriorityQueue[ScheduledItem[typing.TState]] = PriorityQueue()
 
         super().__init__()
 
@@ -78,7 +78,7 @@ class VirtualTimeScheduler(SchedulerBase):
 
         spinning = 0
         while self.is_enabled:
-            item = self.get_next()
+            item: ScheduledItem[typing.TState] = self.get_next()
             if not item:
                 break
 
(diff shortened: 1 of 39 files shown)


erikkemperman:
Hmm I just noticed that the little benchmark I included behaves differently on Travis for PRs than regular push builds. That's fixed now, and I've also added a warm-up round to avoid potential bias due to one-time startup artifacts.

Here is the summary:

| Platform | Master | PR     | Link |
|----------|--------|--------|------|
| Linux    | 6.06 s | 5.79 s | [here]  |
| MacOS    | 7.94 s | 7.55 s | [here]  |
| Windows  | 9.64 s | 6.36 s | [here]  |

[1]: https://travis-ci.org/ReactiveX/RxPY/jobs/507137635#L1094
[2]: https://travis-ci.org/ReactiveX/RxPY/jobs/507137637#L991
[3]: https://travis-ci.org/ReactiveX/RxPY/jobs/507137638#L948


I will leave this PR alone now for a while, promise!

erikkemperman:
Thanks for reviewing, I realize it’s a lot... So the workflow now is it needs one more approval? I see you assigned this to me, does that mean I should merge my own PR if another approval is marked?

jcafhe:
Your detailed abstract above is much much appreciated :+1: . I like the way you simplify `PriorityQueue` by removing synchronization. 

> - Also, the CurrentThreadScheduler uses a dict to make sure that each invoking thread is mapped to its own queue. But that is just what threading.local does, no need to implement it manually (and leaking keys in the dict, as threads disappear).
> - Refactoring to address the above, we find that there is no longer a need for re-entrant locks and we can use the cheaper regular lock.

I've seen that you have completely remove explicit reference to lock in `CurrentThreadScheduler`. I suppose that everything is done by `threading.local` when en-queuing but I'm not 100% sure, so maybe if you could confirm:
https://github.com/ReactiveX/RxPY/blob/4a924287dc6dc3bb1c7d1199b4b4d9f83776bb89/rx/concurrency/currentthreadscheduler.py#L117-L118

Also the use of  `__new__`  is cool too :+1:

erikkemperman:
@jcafhe Thanks for taking the time to review! Yeah, the tread local means that each thread will have exclusive access to its own private queue, therefore there is no need for explicit locking — other threads never see the same queue.

dbrattli:
Cool fix for the CurrentThreadScheduler, and yes with enough approvals you are free to merge yourself.

erikkemperman:
@dbrattli Thanks. From earlier PR, it seems that the preferred method of merging into master is to squash the commits into a single one? That’s fine by me, especially as there are so many here separated mostly for review purposes — but I was just wondering if it has implications for ability to revert individual changes easily, if needed later?

rx/concurrency/virtualtimescheduler.py
```diff
@@ -12,129 +14,155 @@
...
         Args:
             initial_clock: Initial value for the clock.
-            comparer: Comparer to determine causality of events based
-                on absolute time.
         """
-        self.clock = initial_clock
-
-        self.is_enabled = False
-        self.queue = PriorityQueue(1024)
+        self._clock = initial_clock
```
jcafhe: Maybe I'm missing something, but making the `clock` member pseudo-private (by convention) seems to be a bit misleading. It's explicitly used elsewhere, e.g. by the `HistoricalScheduler` or in tests. What do you think?

erikkemperman: Ah, good point! I made it private because setting this clock from unsynchronized code will cause problems, and figured it would be alright for “trusted” code like HistoricalScheduler and tests to access it.

But you’re right, of course, it would have been better to add a public read-only property here. I will do that tonight and push to this PR.

Thanks!!

jcafhe: :+1: I will still approve this PR so you should be able to merge, since I may be not available tomorrow. Great Job!

erikkemperman: @jcafhe I just pushed a change that addresses your point. Does that look all right? Thanks again for the thorough review, this is a big PR and it makes me feel slightly more secure knowing that multiple eyes went over it!

erikkemperman: Hmm, sorry, I've had to force-push that last bit twice -- I had a little change in my local working copy and that got included by accident. Fixed now.

jcafhe: yes that feels good for me :+1: 
> this is a big PR and it makes me feel slightly more secure knowing that multiple eyes went over it!

I must admit I've struggled with the run loop in `EventLoopScheduler`, I didn't manage to fully understand what's going on.

erikkemperman: Glad that you mention that, it means the code change isn't self-explanatory and so I should have been more verbose about the changes in EventLoopScheduler, besides the synchronization of its queue.

Currently, waiting until the up-next item is due looks like this:
- EventLoopScheduler starts a separate Timer and blocks, waiting for its threading.Condition to be notified. 
- The timer is just a secondary thread which has a threading.Event on which it waits with a timeout, unless the event is set (cancelled).
- After timeout, this secondary thread then invokes a callback, which in this case is the tick method of our scheduler.
- Finally, the tick method notifies the original threading.Condition and our main loop resumes.

So it occurred to me that the main loop, instead of doing an indefinite wait for an outside timer, ought to simply add a timeout argument to its own initial wait.

Furthermore, while considering this code, it seemed to me that it was possible for edge cases to go out-of-order, due to the fact that there are two routes for an item to being invoked (queue and ready_list). So I added a little bit of logic that makes sure those are merged / zipped correctly into each iteration's actual work order.

Does this help?

jcafhe: Thank you for your deep insight, it is very helpful. I understand now.
Just to be sure, I understand that there's a sort by duetime between items from `_queue` and items from the `_ready_list` before appending to the `ready` deque. To be more accurate, the comparison is done with the first item in `_ready_list` so it is assumed that `_ready_list` is already sorted in some way.

jcafhe: Forget my comment, `_ready_list` should be de facto sorted since it's only used to schedule ASAP.

erikkemperman: Yes, that’s exactly right; we can do a merge here rather than a full sort, since ready_list and the queue are each already sorted.

So, do you think the PR is ready to be merged?

jcafhe: Yes sure.

erikkemperman: Excellent, thanks again for the detailed review!

PR #329, opened 2019-03-17
Simplify coveralls configuration (zero); fixes #316

### LINKED ISSUE #316: Coveralls reports

erikkemperman:
I was wondering if just removing custom Coveralls configuration and instead relying on automatic / defaults would fix #316. Looks like it does?

(Worked hard at the PQ stuff, would love to see a report with a little coverage bump :-) Comparing branches locally is tedious).

### DIFF
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,2 +1,0 @@
-service_name: travis-pro
-repo_token: XXYpw6bql9Q6dtYESB2iMIUkeudK3Kiyy
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ matrix:
       python: 3.6
       # Coverage for the baseline 3.6 build: include some optional packages
       before_script:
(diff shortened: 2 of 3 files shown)


erikkemperman:
Well coveralls is already reporting correctly for this PR, so yeah I think this was it!

Funny that it worked for some builds though — maybe those were directly pushed rather than merged from a PR?

erikkemperman:
Thanks for review guys, I’ll merge this now!

erikkemperman:
Yay, it works for the merge commits as well.

PR #330, opened 2019-03-17
Doc: fixed docstrings and typing

MainRo:
This PR fixes doc strings comments that generate warnings in sphinx.
I also changed the typing of *ScheduledAction* in *rx/core/typing.py*. I suppose that this was a typo ?


MainRo:
ok, the quotes in Schedulers indicate a forward reference. I must see how to get sphinx working with this.

PR #335, opened 2019-04-01
Add badge for build status at ReadTheDocs

dbrattli:


README.rst
```diff
@@ -10,6 +10,10 @@ The Reactive Extensions for Python (RxPY)
 
 .. image:: https://img.shields.io/pypi/v/rx.svg
         :target: https://pypi.python.org/pypi/Rx
+        
+.. image:: https://readthedocs.org/projects/rxpy/badge/?version=latest
```
erikkemperman: To be consistent with the other badges, this could also use the img.shields.io service?

https://img.shields.io/readthedocs/:packageName.svg

erikkemperman: Or, including the version,
https://img.shields.io/readthedocs/:packageName/:version.svg

dbrattli: Thanks, should be fixed now

erikkemperman: would it be better to point at the later docs (from the readme for master, after all)?

```
.. image:: https://img.shields.io/readthedocs/rxpy/latest.svg
        :target: https://rxpy.readthedocs.io/en/latest/
        :alt: Documentation Status
```

Doesn’t matter much, I guess.

dbrattli:
Considered linking to the docs, but for me the badge is about build status, and when builds are failing I want to click and see the builds, not the docs.

erikkemperman:
Ah yes, that makes sense.

Would be nice if there was a link to the docs from pypi, though, for the appropriate version?

MainRo:
The links in the readme should be fixed in #340. We can also add a link on the left side of pypi with an additional paramater in setup:

```
project_urls={
        'Docs: RTD': 'https://rxpy.readthedocs.io'
},
```

erikkemperman:
I think that that could work, replacing the version in setup.py from a variable if it is a “release” value, otherwise just link to latest?

I don’t think that’s possible in the readme, so your proposal in the sidebar sounds good to me!

PR #340, opened 2019-04-01
Fixed broken documentation link

bradleynull:
Fixed broken documentation link.


dbrattli:
Thanks, but could we point to "Read the Docs" instead, i.e https://rxpy.readthedocs.io/en/latest/ and also fix the other two links below the first one as well!?

bradleynull:
@dbrattli Fixed the links per your request! :)

PR #344, opened 2019-04-03
Remove nonexist

erikkemperman:
I think these were left over from the refactoring of Observable recently, can they simply be removed?


dbrattli:
Adding @thomasnield that is the author of the statistics methods to check if he wants to refactor them into RxPY 3.  The statistics functions could  be made into a separate package now that the "extension" methods are gone and operators are just plain pipeable functions.

erikkemperman:
Ah, I had not even noticed that module was broken besides the non-existent import... Shouldn't be too hard to refactor I would guess. Maybe it would also be nice to keep running stats, as opposed to just on completion.

dbrattli:
Approving this for now, so you can decide if you want to refactor the code in a separete PR (or not).

erikkemperman:
I've made a mental note to revisit this later, but I think I will merge this part now.

PR #349, opened 2019-04-16
Repair some leftovers of millisecond to second change

erikkemperman:
I found a few cases where, it seems to me, there were some leftovers from the old millisecond arguments.

### DIFF
--- a/rx/__init__.py
+++ b/rx/__init__.py
@@ -337,7 +337,7 @@ def generate_with_relative_time(initial_state, condition, iterate, time_mapper) 
     initial state until the condition fails.
 
     Example:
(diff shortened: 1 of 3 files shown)


jcafhe:
As a bonus, maybe you could complete the `generate_with_relative_time` signature by specifying types in *rx/\_\_init__.py*.

erikkemperman:
@jcafhe Yes, actually I was planning (hoping) to attempt a big push in typing, perhaps even getting mypy into CI. I would prefer to keep this separate, though, does that make sense?

jcafhe:
No problem, as you wish :+1:

erikkemperman:
Thanks for quick review!

PR #350, opened 2019-04-16
Fix GtkScheduler and its tests

erikkemperman:
I fixed an error in the GtkScheduler, but at first I did not understand why this was not caught by the unittests earlier...

A little bit weird: it might be dangerous to do assertions from within callbacks -- at least in this case of the GtkScheduler, it seems that the thread which invokes the callbacks eats exceptions, so they never actually cause the unit tests to fail...

I think some of the other tests might have to be visited to check for this phenomenon as well, hopefully I might find some time to do that the next couple of days.

### DIFF
--- a/rx/concurrency/mainloopscheduler/gtkscheduler.py
+++ b/rx/concurrency/mainloopscheduler/gtkscheduler.py
@@ -57,7 +57,7 @@ class GtkScheduler(SchedulerBase):
             The disposable object used to cancel the scheduled action
             (best effort).
         """
-        return self._gtk_schedule(duetime, action, state)
+        return self._gtk_schedule(duetime, action, state=state)
 
     def schedule_absolute(self, duetime: typing.AbsoluteTime, action: typing.ScheduledAction,
                           state: typing.TState = None) -> typing.Disposable:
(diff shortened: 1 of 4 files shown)


jcafhe:
> A little bit weird: it might be dangerous to do assertions from within callbacks -- at least in this case of the GtkScheduler, it seems that the thread which invokes the callbacks eats exceptions, so they never actually cause the unit tests to fail...

I know that PyQt sometimes  eats exceptions too (can't remember in which case exactly, I believe this is related to signals/slots).

erikkemperman:
I did find some more issues around assertions within callbacks, but haven't figured out the fixes yet. I just pushed some work in progress to this branch by accident, fixed that now. But perhaps this PR should remain open for a bit.

erikkemperman:
OK I think this remedies the situation for the concurrency package, haven't checked anything else for this though.

To reiterate, it turns out to be slightly dangerous for "3rd party threads" to be executing assertions within the unit tests. They could eat the exception, which would mean the tests pass when they should not.

Furthermore, if we rely on the callbacks to signal back completion, i.e. via a semaphore or deferred, uncaught errors might prevent that from happening and we have blocked tests.

erikkemperman:
Thanks for review!

PR #361, opened 2019-04-27
SchedulerBase polish, typings, test

erikkemperman:
I am once again trying to separate moderately sized bits off a branch with way too many changes in it...

### DIFF
--- a/rx/concurrency/schedulerbase.py
+++ b/rx/concurrency/schedulerbase.py
@@ -6,6 +6,7 @@ from rx.disposable import Disposable, MultipleAssignmentDisposable
 from rx.internal.basic import default_now
 from rx.internal.constants import DELTA_ZERO, UTC_ZERO
 
+
 class SchedulerBase(typing.Scheduler):
     """Provides a set of static properties to access commonly used
     schedulers.
@@ -15,6 +16,17 @@ class SchedulerBase(typing.Scheduler):
                       action: typing.ScheduledAction,
                       state: Optional[typing.TState] = None
                       ) -> typing.Disposable:
+        """Invoke the given given action. This is typically called by instances
+        of ScheduledItem.
+
+        Args:
+            action: Action to be executed.
+            state: [Optional] state to be given to the action function.
+
+        Returns:
+            The disposable object returned by the action, if any; or a new
+            (no-op) disposable otherwise.
+        """
         ret = action(self, state)
         if isinstance(ret, typing.Disposable):
             return ret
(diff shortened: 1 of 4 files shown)


rx/concurrency/schedulerbase.py
```diff
@@ -6,15 +7,94 @@
...
+        property.
+
+        Returns:
+             The scheduler's current time, as a datetime instance.
+         """
+
+        return default_now()
+
+    @abstractmethod
+    def schedule(self,
```
jcafhe: I'm just curious, is it worth to re-declare `schedule`, `schedule_relative` & `schedule_absolute` as abstract methods while these are already declared in `typing.Scheduler` ?

rx/concurrency/schedulerbase.py
```diff
@@ -77,57 +157,86 @@ def now(self) -> datetime:
...
+        elif not isinstance(value, datetime):
+            value = datetime.utcfromtimestamp(value)
+
+        return value
 
     @classmethod
-    def to_timedelta(cls, timespan: typing.AbsoluteOrRelativeTime) -> timedelta:
-        """Converts time value to timedelta"""
+    def to_timedelta(cls, value: typing.AbsoluteOrRelativeTime) -> timedelta:
+        """Converts time value to timedelta. This method handles both absolute
```
jcafhe: Maybe we could specify in the doc that the returned timedelta is relative to POSIX time 0 (January 1, 1970 00:00)

rx/concurrency/schedulerbase.py
erikkemperman: Yeah, maybe not strictly necessary... Although it might help to keep the generated docs and `help()` complete.

I guess I included this mainly because of an idea I have that I’m playing around with, which would simplify all of the classes that extend `SchedulerBase` by putting some often-repeated logic (schedule methods being defined in terms of one another) in the base class.

Having these here at this point would make that second step a lot easier / readable.

rx/concurrency/schedulerbase.py
erikkemperman: Yes, good catch, that would be useful to mention! I’ll leave this PR alone for a bit, not to complicate its review, but I’ll add that later!

erikkemperman:
Ugh I just noticed I accidentally included a change in virtualtime scheduler here... It should have been in that other PR.

Sorry about that, I’ve been juggling branches and commits and this fell through. Please ignore that part, I’ll transplant it later.

rx/concurrency/schedulerbase.py
jcafhe: cool :+1:

tests/test_concurrency/test_virtualtimescheduler.py
```diff
@@ -29,28 +29,38 @@ def action(scheduler, state):
         assert(ran[0])
 
     def test_virtual_schedule_action_error(self):
-        ex = 'ex'
+        msg = 'ex'
+        exc = None
         try:
```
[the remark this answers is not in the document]
erikkemperman: Hmm, I did not know about that trick. Thanks! Yes, agree that might be better. But I’ll have to change it in a lot of different places... Maybe I’ll find some time tomorrow.

erikkemperman:
FYI, I removed the bit which was not supposed to be in this PR, and I added a sentence about `to_timedelta` being relative to the epoch (1970-01-01) as suggested by @jcafhe

rx/concurrency/schedulerbase.py
erikkemperman: I've added a sentence mentioning this, thanks!

tests/test_concurrency/test_virtualtimescheduler.py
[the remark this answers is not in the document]
erikkemperman: I've moved this bit to the other PR, where it belongs. But thanks, I used your suggestion there!

erikkemperman:
Thanks for review!

PR #362, opened 2019-04-27
Scheduler docstrings, typing, polish

erikkemperman:
Sorry, this is a pretty huge diff. But I tried to make sure that these changes are all basically trivial.

### DIFF
--- a/tests/test_concurrency/test_threadpoolscheduler.py
+++ b/tests/test_concurrency/test_threadpoolscheduler.py
@@ -18,6 +18,13 @@ class TestThreadPoolScheduler(unittest.TestCase):
         diff = scheduler.now - default_now()
         assert abs(diff) < timedelta(milliseconds=1)
 
+    def test_threadpool_now_units(self):
+        scheduler = ThreadPoolScheduler()
+        diff = scheduler.now
+        sleep(0.1)
+        diff = scheduler.now - diff
+        assert timedelta(milliseconds=80) < diff < timedelta(milliseconds=180)
+
     def test_schedule_action(self):
         ident = threading.current_thread().ident
         evt = threading.Event()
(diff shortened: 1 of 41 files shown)


tests/test_concurrency/test_threadpoolscheduler.py
```diff
@@ -56,22 +67,22 @@ def test_schedule_action_absolute(self):
...
             evt.set()
 
-        nt.schedule_absolute(datetime.utcnow()+timedelta(milliseconds=100), action)
+        nt.schedule_absolute(default_now() + timedelta(milliseconds=100), action)
         evt.wait()
 
     def test_schedule_action_cancel(self):
         nt = thread_pool_scheduler
-        set = []
+        set = False
```
jcafhe: Maybe it would be more relevant to rename `set` to  `ran`?

jcafhe:
Huge diff indeed! 

It may be not the place to discuss this here, but I was wondering if we should take care of one corner case  in `schedule_absolute` or `schedule_relative`.  Basically, we can say that if duetime equals 0.0s, `schedule` can be used as a 'shortcut'. 
Should it be `duetime <= 0.0` instead, i.e. do you think duetime can be negative in some cases?

erikkemperman:
@jcafhe Thanks for thorough review, extremely helpful!

I followed almost all of your suggestions, except the symmetric time diffs in the unit tests: I found that, especially on Travis, we have to allow a bit more room on the "right-hand side", as their VMs are competing for physical hardware.

I also agree about the "shortcut" options but am planning to address that separately (which I mentioned earlier, consolidating some login in SchedulerBase).

jcafhe:
>  I found that, especially on Travis, we have to allow a bit more room on the "right-hand side", as their VMs are competing for physical resources.

Got it :+1: 

Good job!

erikkemperman:
Thanks for review!

Issue #358, opened 2019-04-23
Enforce CurrentThreadScheduler to run on a dedicated thread?


MichaelSchneeberger:
I have two concerns about the `rx.concurrency.currentthreadscheduler.CurrentThreadScheduler` class implementation. Both relate to the `__new__` method:

``` python
    def __new__(cls) -> 'CurrentThreadScheduler':
        """Ensure that each thread has at most a single instance."""

        thread = threading.current_thread()
        self: 'CurrentThreadScheduler' = CurrentThreadScheduler._global.get(thread)
        if not self:
            self = super().__new__(cls)
            CurrentThreadScheduler._global[thread] = self
        return self
```

**1.** Is a `CurrentThreadScheduler` always supposed to schedule actions on the same thread? In my opinion, the `CurrentThreadScheduler` is attached to some thread once the trampoline is run for the first time or "reactivated".

**2.** When I tried to extend from `CurrentThreadScheduler` in [rxbackpressure](https://github.com/MichaelSchneeberger/rxbackpressure) I couldn't do it because the `__new__` method would always return a `CurrentThreadScheduler`.

``` python
class MyCurrentThreadScheduler(SchedulerMixin, CurrentThreadScheduler):
    pass

my_current_thread_scheduler = MyCurrentThreadScheduler()
print(type(my_current_thread_scheduler))     # would print CurrentThreadScheduler
```

erikkemperman:
My understanding of the semantics of CurrentThreadScheduler is that the same thread which invokes the schedule method should also invoke the scheduled callback.

This is independent of which thread _created_ the scheduler. So even if you create a scheduler instance on thread A and pass it to thread B, if B calls schedule then the callback will be run on B.

Perhaps this behavior is not intuitive? In that case it should be either documented more thoroughly or reconsidered — I can imagine semantics where the creating thread is always used to run callbacks, for instance.

I don’t think my refactoring of this class changed these semantics, by the way, or at least it wasn’t supposed to... And it will remain true if you remove the __new__ implementation.

The latter is only to cover the common case where the same thread which creates the scheduler instance also does the scheduling: since each thread will have its own queue and trampoline, it does not make sense imho to have multiple scheduler instances per thread.

I admit I did not consider extending the class though... I suppose you’d have to override the __new__ method as well? I’ll give that some thought!

dbrattli:
[...]

PR #363, opened 2019-04-28
Don't instantiate new CurrentThreadScheduler._Local with each scheduler instance

erikkemperman:
This fixes the problem pointed out by @MichaelSchneeberger in https://github.com/ReactiveX/RxPY/issues/358.

### DIFF
--- a/rx/concurrency/currentthreadscheduler.py
+++ b/rx/concurrency/currentthreadscheduler.py
@@ -47,6 +47,7 @@ class CurrentThreadScheduler(SchedulerBase):
     that will block the current thread while waiting.
     """
 
+    _local = _Local()
     _global: MutableMapping[threading.Thread, 'CurrentThreadScheduler'] = WeakKeyDictionary()
 
     def __new__(cls) -> 'CurrentThreadScheduler':
@@ -62,9 +63,7 @@ class CurrentThreadScheduler(SchedulerBase):
     def __init__(self) -> None:
         """Creates a scheduler that schedules work as soon as possible
         on the current thread."""
-
         super().__init__()
-        self._local = _Local()
 
     def schedule(self,
                  action: typing.ScheduledAction,
@@ -127,7 +126,7 @@ class CurrentThreadScheduler(SchedulerBase):
 
         si: ScheduledItem[typing.TState] = ScheduledItem(self, state, action, duetime)
 
-        local: _Local = self._local
+        local: _Local = CurrentThreadScheduler._local
         local.queue.enqueue(si)
         if local.idle:
             local.idle = False
@@ -147,7 +146,7 @@ class CurrentThreadScheduler(SchedulerBase):
         False; otherwise, if the trampoline is not active, then it
         returns True.
         """
-        return self._local.idle
+        return CurrentThreadScheduler._local.idle
 
     def ensure_trampoline(self, action):
         """Method for testing the CurrentThreadScheduler."""
--- a/rx/concurrency/currentthreadscheduler.py
+++ b/rx/concurrency/currentthreadscheduler.py
@@ -45,6 +45,15 @@ class CurrentThreadScheduler(SchedulerBase):
     """Represents an object that schedules units of work on the current thread.
     You never want to schedule timeouts using the CurrentThreadScheduler since
     that will block the current thread while waiting.
+
+    Please note, there will be at most a single instance per thread -- calls to
+    the constructor will just return the same instance if one already exists.
+
+    Conversely, if you pass an instance to another thread, it will effectively
+    behave as a separate scheduler, with its own queue. In particular, this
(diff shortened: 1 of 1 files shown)


MichaelSchneeberger:
Do you actually still need the `_global` field and the `__new__` method?

erikkemperman:
@MichaelSchneeberger Well, only if it is decided that we want to enforce, not just internally but for end-users as well, that there is at most a single instance of `CurrentThreadScheduler` per thread. Unless, do you know a better way to enforce this?

erikkemperman:
I've added a note which I hope clears up some of the not-so-intuitive behaviour of the CurrentThreadScheduler, as it currently stands.

@MichaelSchneeberger Do you think this is sufficiently clear?
https://github.com/ReactiveX/RxPY/pull/363/commits/b782765718b875e468ab1a2d56b11931f48972ba

MichaelSchneeberger:
the best in my opinion would be to have backwards compatibility with the CurrentThreadScheduler v1.6, which allows to have nested CurrentThreadSchedulers as the factorial example in #358 illustrates.

A singleton behavior could then be added by introducing a classmethod `singleton` for instance.

``` python
scheduler = CurrentThreadScheduler.singleton()
```

Another point I mentioned before is that the current implementation of `CurrentThreadScheduler` also works without the `__new__` method and `_global` field.

``` python
class CurrentThreadScheduler(SchedulerBase):
    """Represents an object that schedules units of work on the current thread.
    You never want to schedule timeouts using the CurrentThreadScheduler since
    that will block the current thread while waiting.
    """
    class Local(threading.local):
        __slots__ = 'idle', 'queue'

        def __init__(self):
            self.idle: bool = True
            self.queue: PriorityQueue[ScheduledItem[typing.TState]] = PriorityQueue()

    _local = Local()

    def __init__(self, local: 'CurrentThreadScheduler.Local' = None):
        """Creates a scheduler that schedules work as soon as possible
        on the current thread."""
        super().__init__()
```

To test the `CurrentThreadScheduler`, I used the following code ...

``` python
scheduler1 = EventLoopScheduler()
scheduler2 = EventLoopScheduler()

scheduler = CurrentThreadScheduler()
print(scheduler._local.queue)
scheduler = CurrentThreadScheduler()
print(scheduler._local.queue)

def show_current_thread_scheduler(_):
    def gen1():
        for _ in range(4):
            scheduler = CurrentThreadScheduler()
            yield scheduler._local.queue
    print(list(gen1()))

rx.just(None, scheduler=scheduler1).pipe(
    rx.operators.map(show_current_thread_scheduler),
    rx.operators.zip(rx.just(None, scheduler=scheduler2).pipe(rx.operators.map(show_current_thread_scheduler))),
).run()
```

... which prints ...

```
<rx.internal.priorityqueue.PriorityQueue object at 0x7f84d8a7e390>
<rx.internal.priorityqueue.PriorityQueue object at 0x7f84d8a7e390>
[<rx.internal.priorityqueue.PriorityQueue object at 0x7f84d89e0e10>, <rx.internal.priorityqueue.PriorityQueue object at 0x7f84d89e0e10>, <rx.internal.priorityqueue.PriorityQueue object at 0x7f84d89e0e10>, <rx.internal.priorityqueue.PriorityQueue object at 0x7f84d89e0e10>]
[<rx.internal.priorityqueue.PriorityQueue object at 0x7f84d89e0898>, <rx.internal.priorityqueue.PriorityQueue object at 0x7f84d89e0898>, <rx.internal.priorityqueue.PriorityQueue object at 0x7f84d89e0898>, <rx.internal.priorityqueue.PriorityQueue object at 0x7f84d89e0898>]
```

erikkemperman:
Well, speaking just for myself, I have a feeling that using "nested CurrentThreadSchedulers" in this fashion, to achieve a kind of synchronization, is perhaps not the prettiest pattern...

I am not sure if scenarios like these ought to be facilitated, and if they do if there isn't a better, cleaner way to accomplish that. My feeling is that it was never actually explicitly intended to work this way -- i.e. that it works like it does in 1.6 more or less by accident. But, since I wasn't around at the time, I might very well be wrong about that!

However, having said all that, you are absolutely correct that the current behaviour of CurrentThreadScheduler is, in this respect, different from 1.x.

Since I am the author of this change, and also new around here, I think I should not weigh in (more than I already have) on what the resolution should be here.

So I'll wait what others might propose?  If it is decided to keep this change, or something like it, I hope the doc I added here might be sufficient in the way of announcing this slight departure from 1.x semantics.

erikkemperman:
I think this should be merged now — the question of @MichaelSchneeberger is not yet resolved, but this PR fixes a legitimate bug (local constructor) and documents the current state of the master branch.

I propose to continue discussion of how to proceed from there in the issue https://github.com/ReactiveX/RxPY/issues/358.

Issue #365, opened 2019-05-03
Wrong take_while_indexed example


Hongbo-Miao:
`take_while_indexed` currently is using `take_while` example.

https://github.com/ReactiveX/RxPY/blob/e92a80f695489fff652a953a530d86d5a931a599/rx/operators/__init__.py#L2828-L2830

Hongbo-Miao:
Fixed by https://github.com/ReactiveX/RxPY/pull/366

PR #366, opened 2019-05-03
fix: take_while_indexed example

Hongbo-Miao:
Fix https://github.com/ReactiveX/RxPY/issues/365


Hongbo-Miao:
@erikkemperman you are right. Just updated. Thanks!

PR #369, opened 2019-05-07
Install optional packages for Python 3.7 on all Travis platforms

erikkemperman:
I thought it would be good to try and install the various optional dependencies on the other Travis platforms as well. This mostly turned out to be reasonably easy, except I haven't yet figured out how to get GTK working on Windows.

Furthermore, I figured that it might be convenient to have a couple of "fast" builds, without the optional dependencies and without that little benchmark at the end.

While I was preparing this PR, running a bunch of Travis builds one after the other, I noticed that sometimes the build fails on some timing assertion -- mostly the MacOS one. Restarting the build then usually works, proving that the code is actually in order.

But this flakiness gets a bit worse with the extra tests that are now being executed due to the optional dependencies being available. So I am also working on relaxing the timings slightly, I'll submit that separately when it's ready.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,51 +1,64 @@
 matrix:
   include:
-    - os: linux
+
+    - name: "Python 3.6 on Linux"
+      python: "3.6"
+      os: linux
       dist: xenial
       language: python
-      python: '3.6'
-      # Coverage for the baseline 3.6 build: include some optional packages
-      before_script:
-        # install some packages to ensure all of the optional tests are covered
-        - pip3 install eventlet gevent pygame pyqt5 pyside2 tornado twisted
-        # pycairo / pygobject need native libraries
+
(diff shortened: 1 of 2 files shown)


dbrattli:
If there is places we could use virtual time instead then we should do that. But for testing the actual schedulers themselves we would need to work in real time.

erikkemperman:
Yeah, the flakiness seems only due to the real time scheduler tests. But I think I can make those robust, we just have to allow for Travis VMs to be a bit sluggish -- the important thing to test is that the scheduled items are run within the expected order of magnitude -- a couple millis more should not fail the build, I think.

Anyway, this is what Travis looks like with this PR: https://travis-ci.org/ReactiveX/RxPY/builds/529133727

erikkemperman:
Sorry, amended to correct one slight mistake, the benchmark was not actually run where I wanted it to.

PR #372, opened 2019-05-10
Fix type hints for pipe using overloads. Fixes #355, #285

### LINKED ISSUE #355: Fix type hints for pipe
The current type hints for the pipe operator is not correct since it assumes that the final type will always be an `Observable`. Thus there's problems when piping to functions such as `to_future` that returns a `Future`.  The current typing hints are:

```py
def pipe(*operators: Callable[[Observable], Observable]) -> Callable[[Observable], Observable]:
```

The correct would be that the callables were `Callable[[A], B]` and `Callable[[B], C]` etc, but this is not possible when using the star operator as above. We would probably need to do something like using overloads and specifying the N first cases as with [RxJS and typescript](https://github.com/ReactiveX/rxjs/blob/a9c87f6a6b7deed12950ac926409a0c4617f4313/src/internal/Observable.ts#L309) e.g:

```ts
/* tslint:disable:max-line-length */
  pipe(): Observable<T>;
  pipe<A>(op1: OperatorFunction<T, A>): Observable<A>;
  pipe<A, B>(op1: OperatorFunction<T, A>, op2: OperatorFunction<A, B>): Observable<B>;
  pipe<A, B, C>(op1: OperatorFunction<T, A>, op2: OperatorFunction<A, B>, op3: OperatorFunction<B, C>): Observable<C>;
  pipe<A, B, C, D>(op1: OperatorFunction<T, A>, op2: OperatorFunction<A, B>, op3: OperatorFunction<B, C>, op4: OperatorFunction<C, D>): Observable<D>;
  pipe<A, B, C, D, E>(op1: OperatorFunction<T, A>, op2: OperatorFunction<A, B>, op3: OperatorFunction<B, C>, op4: OperatorFunction<C, D>, op5: OperatorFunction<D, E>): Observable<E>;
  pipe<A, B, C, D, E, F>(op1: OperatorFunction<T, A>, op2: OperatorFunction<A, B>, op3: OperatorFunction<B, C>, op4: OperatorFunction<C, D>, op5: OperatorFunction<D, E>, op6: OperatorFunction<E, F>): Observable<F>;
  pipe<A, B, C, D, E, F, G>(op1: OperatorFunction<T, A>, op2: OperatorFunction<A, B>, op3: OperatorFunction<B, C>, op4: OperatorFunction<C, D>, op5: OperatorFunction<D, E>, op6: OperatorFunction<E, F>, op7: OperatorFunction<F, G>): Observable<G>;
  pipe<A, B, C, D, E, F, G, H>(op1: OperatorFunction<T, A>, op2: OperatorFunction<A, B>, op3: OperatorFunction<B, C>, op4: OperatorFunction<C, D>, op5: OperatorFunction<D, E>, op6: OperatorFunction<E, F>, op7: OperatorFunction<F, G>, op8: OperatorFunction<G, H>): Observable<H>;
  pipe<A, B, C, D, E, F, G, H, I>(op1: OperatorFunction<T, A>, op2: OperatorFunction<A, B>, op3: OperatorFunction<B, C>, op4: OperatorFunction<C, D>, op5: OperatorFunction<D, E>, op6: OperatorFunction<E, F>, op7: OperatorFunction<F, G>, op8: OperatorFunction<G, H>, op9: OperatorFunction<H, I>): Observable<I>;
  pipe<A, B, C, D, E, F, G, H, I>(op1: OperatorFunction<T, A>, op2: OperatorFunction<A, B>, op3: OperatorFunction<B, C>, op4: OperatorFunction<C, D>, op5: OperatorFunction<D, E>, op6: OperatorFunction<E, F>, op7: OperatorFunction<F, G>, op8: OperatorFunction<G, H>, op9: OperatorFunction<H, I>, ...operations: OperatorFunction<any, any>[]): Observable<{}>;
```

But I'm unsure how to do this properly in Python so that IDEs, mypy and linters acually can handle this so you get proper completion on the stuff that comes out of the pipe.

dbrattli:
This PR fixes the typing issues with the `pipe` operator using typing overloads. The downside is that pylint does not support this yet and only sees the first overload (https://github.com/PyCQA/pylint/issues/2239). The good thing is that PyCharm does the right thing and see all the overloads and correctly infers the last type.

Skjermbilde 2019-05-10 kl  17 40 30

Thus I would rather do it the right way since PyCharm already supports this, and let pylint problems be pylint problems. This is similary to how the typing were fixed with typescript in RxJS (https://github.com/ReactiveX/rxjs/blob/master/src/internal/Observable.ts). Hopefully pylint will eventually be fixed.

### DIFF
--- a/rx/core/observable/observable.py
+++ b/rx/core/observable/observable.py
@@ -216,7 +216,25 @@ class Observable(typing.Observable):
         return Disposable(auto_detach_observer.dispose)
 
     @overload
-    def pipe(self) -> 'Observable':  # pylint: disable=no-self-use
+    def pipe(self, *operators: Callable[['Observable'], 'Observable']) -> 'Observable':  # pylint: disable=no-self-use
+        """Compose multiple operators left to right.
+
+        Composes zero or more operators into a functional composition.
+        The operators are composed to right. A composition of zero
+        operators gives back the original source.
+
+        source.pipe() == source
+        source.pipe(f) == f(source)
+        source.pipe(g, f) == f(g(source))
+        source.pipe(h, g, f) == f(g(h(source)))
+        ...
+
+        Returns the composed observable.
+        """
+        ...
+
+    @overload
+    def pipe(self) -> 'Observable':  # pylint: disable=function-redefined, no-self-use
         ...  # pylint: disable=pointless-statement
 
     @overload
@@ -264,7 +282,7 @@ class Observable(typing.Observable):
         ...  # pylint: disable=pointless-statement
 
     @overload
-    def pipe(self,  # pylint: disable=function-redefined,too-many-arguments,no-self-use
+    def pipe(self,  # pylint: disable=function-redefined, no-self-use, too-many-arguments
              op1: Callable[['Observable'], A],
              op2: Callable[[A], B],
              op3: Callable[[B], C],
@@ -275,7 +293,7 @@ class Observable(typing.Observable):
         ...  # pylint: disable=pointless-statement
 
     # pylint: disable=function-redefined
-    def pipe(self, *operators: Callable[['Observable'], Any]) -> Any:  # type: ignore
+    def pipe(self, *operators: Callable[['Observable'], Any]) -> Any:
         """Compose multiple operators left to right.
 
         Composes zero or more operators into a functional composition.
--- a/rx/core/observable/observable.py
+++ b/rx/core/observable/observable.py
@@ -217,61 +217,61 @@ class Observable(typing.Observable):
 
     @overload
     def pipe(self) -> 'Observable':  # pylint: disable=no-self-use
-        ... # pylint: disable=pointless-statement
-
-    @overload
-    def pipe(self, op1: Callable[['Observable'], A]) -> A:  # pylint: disable=function-redefined, no-self-use
-        ... # pylint: disable=pointless-statement
-
-    @overload
-    def pipe(self,
-             op1: Callable[['Observable'], A],
-             op2: Callable[[A], B]) -> B:  # pylint: disable=function-redefined, no-self-use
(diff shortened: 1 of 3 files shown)


rx/core/observable/observable.py
```diff
@@ -208,8 +215,67 @@ def set_disposable(_: abc.Scheduler = None, __: Any = None):
...
+
+    @overload
+    def pipe(self,
+             op1: Callable[['Observable'], A],
+             op2: Callable[[A], B],
+             op3: Callable[[B], C],
+             op4: Callable[[C], D],
+             op5: Callable[[D], E],
+             op6: Callable[[E], F],
+             op7: Callable[[E], G]) -> G:  # pylint: disable=function-redefined, no-self-use
```
jcafhe: I believe it should be `F` instead of `E`, i.e. 
```python
op6: Callable[[E], F],
op7: Callable[[F], G]) -> G:
```

dbrattli:
I really don't like that pylint handles this code so badly. I will try to see if I can disable it and only use mypy or try some other linter. Here is another pylint issue that tracks this problem https://github.com/PyCQA/pylint/issues/1581

rx/core/observable/observable.py
dbrattli: Thanks for spotting the typo (since there's really no help in the typing system). Go figure ....

dbrattli:
The latest change with having an Observable specific `pipe()` overload first seems to make pylint happy, so now pylint, mypy and pyre seems to be happy. Adding docstring to the first overload make vscode happy when you inpect or auto-complete.

dbrattli:
We should probably add `py.typed`. Didn't know about it so thanks for the link.

I btw started using Pyre today. Very nice type checking and integration with vscode: https://pyre-check.org/

PR #378, opened 2019-05-15
Fix wrong operator type annotations 

### LINKED ISSUE #377: Invalid publish() annotation?
Hi,

`publish` method is annotated as returning `ConnectableObservable` however - as far as I understand - it actually returns `Callable[[Observable], ConnectableObservable]`. This inconsistency makes Pycharm typechecker go nuts and I have to use `typing.cast` to silence it. Is it my misunderstanding or should I make PR to fix the annotation?

jcafhe:
This PR fixes wrong return type for operators:

- ignore_elements
- publish 
- single_or_default_async
- single_or_default
- skip_last 
- zip_with_iterable

This should also fix #377.

### DIFF
--- a/rx/core/operators/publish.py
+++ b/rx/core/operators/publish.py
@@ -5,7 +5,7 @@ from rx.core import Observable, ConnectableObservable, pipe
 from rx.subjects import Subject
 
 
-def _publish(mapper=None) -> ConnectableObservable:
+def _publish(mapper=None) -> Callable[[Observable], ConnectableObservable]:
     """Returns an observable sequence that is the result of invoking the
     mapper on a connectable observable sequence that shares a single
     subscription to the underlying sequence. This operator is a
--- a/rx/core/operators/publish.py
+++ b/rx/core/operators/publish.py
@@ -1,11 +1,12 @@
-from typing import Callable
+from typing import Callable, Optional
 
 from rx import operators as ops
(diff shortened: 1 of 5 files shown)


rx/core/operators/publish.py
```diff
@@ -5,7 +5,7 @@
 from rx.subjects import Subject
 
 
-def _publish(mapper=None) -> ConnectableObservable:
+def _publish(mapper=None) -> Callable[[Observable], ConnectableObservable]:
```
erikkemperman: Maybe add typing for the mapper argument as well? I'll approve anyway, just a thought.

rx/operators/__init__.py
```diff
@@ -1781,7 +1781,7 @@ def pluck_attr(prop: str) -> Callable[[Observable], Observable]:
     return _pluck_attr(prop)
 
 
-def publish(mapper=None) -> ConnectableObservable:
+def publish(mapper=None) -> Callable[[Observable], ConnectableObservable]:
```
erikkemperman: Same here?

rx/core/operators/publish.py
jcafhe: yes you're right, will fix it, thanks!

jcafhe:
Thank's for reviewing!
As advised by @erikkemperman, I've add type annotations for publish operator parameters too.

PR #382, opened 2019-05-18
add missing type annotations (operators)

jcafhe:
This PR adds type annotations to multiple operators (almost half of them). Hidden functions (implementations files) have been typed too, since it's easier to determine the right signature from the source code.

I've avoided some operators with complex signature like `buffer` & `window`. It should be easier to review them in a separate PR. 

### Note:
`Comparer` type has been added in rx.core.typing defined as `Comparer = Callable[[T1, T2], bool]`

Parameters which are optional, i.e. with a default value of None (except for value of type `Any`) has been systematically typed as `Optional`. I'm not sure if it's required in some special cases, but I guess it can't be harmful.  

### Operators:
- repeat
- observe_on
- multicast
- min & min_by
- merge
- max & max_by
- map & map_indexed
- last_or_default
- last
- join
- group_by_until
- group_by
- first
- distinct_until_changed
- distinct
- group_join
- contains
- average


jcafhe:
Thank's for reviewing!

Issue #381, opened 2019-05-18
provide version number in module.


honglei:
Some thing like:

```python
import rx
rx.__version__
```
[https://stackoverflow.com/questions/20180543/how-to-check-version-of-python-modules](https://stackoverflow.com/questions/20180543/how-to-check-version-of-python-modules)

dbrattli:
Fixed in #384

PR #384, opened 2019-05-18
Adding __version__, fix for issue #381

nandahkrishna:
Fixes issue #381 by adding __version__ to init.py, using code similar to the docs. Please do tell me if there's something I have to change here.

### DIFF
--- a/rx/__init__.py
+++ b/rx/__init__.py
@@ -1,25 +1,11 @@
 # pylint: disable=too-many-lines,redefined-outer-name,redefined-builtin
 
-import os
-import sys
-import re
-from configparser import ConfigParser
-
 from asyncio.futures import Future as _Future
 from typing import Iterable, Callable, Any, Optional, Union
 
 from .core import Observable, abc, typing, pipe
 
-root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
-sys.path.insert(0, root)
-
-with open(os.path.join(root, 'project.cfg')) as project_file:
-    config = ConfigParser()
-    config.read_file(project_file)
-    project_meta = dict(config.items('project'))
-
-__release__ = project_meta['version']
-__version__ = re.sub('[^0-9.].*$', '', __release__)
+__version__ = "3.0.0-beta1"
 
 def amb(*sources: Observable) -> Observable:
     """Propagates the observable sequence that reacts first.
--- a/rx/__init__.py
+++ b/rx/__init__.py
@@ -3,8 +3,19 @@
 from asyncio.futures import Future as _Future
 from typing import Iterable, Callable, Any, Optional, Union
 
+import os
+import re
+from configparser import ConfigParser
+
 from .core import Observable, abc, typing, pipe
 
+with open(os.path.join(root, 'project.cfg')) as project_file:
+    config = ConfigParser()
+    config.read_file(project_file)
+    project_meta = dict(config.items('project'))
+
+__release__ = project_meta['version']
+__version__ = re.sub('[^0-9.].*$', '', __release__)
 
 def amb(*sources: Observable) -> Observable:
     """Propagates the observable sequence that reacts first.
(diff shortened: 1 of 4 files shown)


nandahkrishna:
The Travis CI shows failure on MacOS with benchmarks and optional dependencies (Build 1270.5), but passes the rest.

rx/__init__.py
```diff
@@ -3,8 +3,23 @@
 from asyncio.futures import Future as _Future
 from typing import Iterable, Callable, Any, Optional, Union
 
+import os
```
jcafhe: Maybe modules from the standard library should be imported first, as suggested by the [python style guide](https://www.python.org/dev/peps/pep-0008/#imports).

jcafhe:
> The Travis CI shows failure on MacOS with benchmarks and optional dependencies (Build 1270.5), but passes the rest.

Indeed, but don't worry, we have an issue with MacOs tests for now.

rx/__init__.py
nandahkrishna: Alright, I'll fix this.

nandahkrishna:
Glad to contribute!

MainRo:
isn't opening, reading and parsing a file when loading the rxpy module only to retrieve the version number overkill ? We do the same at different locations (setup.py and doc/conf.py) but not during runtime. I wonder if using bumpversion would not be easier to get the version number in all these locations ?

https://pypi.org/project/bumpversion/

nandahkrishna:
>  I wonder if using bumpversion would not be easier to get the version number in all these locations ?
Certainly does seem like a nice way to go about this. Looking at it, I guess a couple of changes here and there would have to be made, like a new bumpversion.cfg and changes in the docs, etc.

I can go ahead and make the change to bump version if you'd prefer that.

erikkemperman:
Agree it might not be something we want to do at runtime. It's nice to have all these properties in one place, which was the point of project.cfg, but not at any cost...

Maybe the version number could be fixed in the `__init__.py` and read from `setup.py` and `docs/conf.py`? Or, indeed, something like the tool you mention would probably work well.

nandahkrishna:
Well, I think the bumpversion tool seems to be a better option, given that it is quite similar to the current project.cfg - lets you have all the information in the same place. I think I'll give that a try.

nandahkrishna:
The project.cfg file has been removed in favour of .bumpversion.cfg which contains the same details or meta-data, but now supports version string updates using `bumpversion` or `bump2version`.

With the current version being `3.0.0-beta1`, the version has been split into 4 parts - major (3), minor(0), patch(0) and release (beta1).

Here are some examples on how the version string update can be done:
`bump2version release` changes it to `3.0.0-beta2` - this is up for review and the options can be modified to suit the required version styling.
`bump2version minor` increments the minor portion (middle number) by 1, and `bump2version major` changes the major portion or first digit of the version.

The bumpversion or bump2version (better maintained) documents can help- understand this clearly.

Please give me any suggestions you have so I can improve/change this. Thanks!

nandahkrishna:
I guess I could change conf.py and setup.py to use the value rather than read from the cfg, if that's preferred. Should I go ahead and change it or is this fine?

MainRo:
> I guess I could change conf.py and setup.py to use the value rather than read from the cfg, if that's preferred. Should I go ahead and change it or is this fine?

Let's see what @erikkemperman @dbrattli and @jcafhe think about this. Keep it as-is for now.

dbrattli:
Merging this for now as I want to get the next beta out. If there are any issues then lets fail forward and fix the problem.

dbrattli:
Note that I got a bug with this PR when doing the new beta release since the package name became 0.0.0 eg

```
-rw-r--r--  1 dbrattli  staff   183K May 20 07:44 Rx-0.0.0-py2.py3-none-any.whl
-rw-r--r--  1 dbrattli  staff    96K May 20 07:44 Rx-0.0.0.tar.gz
```

I think version needs to be set to build sucessfully. Fixed this by setting `version` = `current_version` in `setup.py`:

```py
version=project_meta.get("current_version"),
```

Could you @nandahkrishna verify that this fix is the preferred way?

nandahkrishna:
Yes, as the .bumpversion.cfg file contains the version number, the line had to be added to setup.py - it is the right solution. Thanks for spotting it, slipped my mind.

PR #385, opened 2019-05-18
fix skip import in PyQt5/PySide2 tests + typos

jcafhe:
When running pytest locally, I'm experiencing some weird behavior with PyQt5 & Pyside2 successive tests and the skip mechanism. I assume this is due to some conflict with the `skip` variable declared at the module level.

Anyway, this PR replaces `skip` variable by `pytest.importorskip` function instead (as in other mainloop scheduler tests), plus replaces 'qt5' by 'pyqt5' to be consistent with 'pyside2'.


jcafhe:
Damn, it fails :disappointed:

erikkemperman:
Yeah, the macos build fails often on the timing tests, I’m working to address that but don’t have a lot of time just now.

The Linux build fails with a core dump though, I think that’s a separate issue which wasn’t there before?

jcafhe:
Yes you are right, the linux build fails and it's not intended. I will try to investigate this.

jcafhe:
@erikkemperman  I've been playing locally with the docker image (what a pain!) which almost matches the travis linux build, i.e. `travisci/ci-sardonyx:packer-1554885359-f909ac5` and I think I've found what's wrong.

In Pyside2 tests,  `app` is a `QCoreApplication` which is supposed to be an application with no GUI. In contrary in PyQt5 tests, `app` is a `QApplication`, i.e. a GUI app which seems to be not supported by the `ci-sardonyx`. Here is what I get during `QApplication` instanciation in a python interactive shell :
```python
python3.7.1) travis@31fa2ffe441b:~/RxPY/tests/test_concurrency/test_mainloopscheduler$ python3
Python 3.7.1 (default, Dec  5 2018, 18:09:53) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from PyQt5.QtWidgets import QApplication
>>> app = QApplication([])
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, wayland-egl, wayland, wayland-xcomposite-egl, wayland-xcomposite-glx, webgl, xcb.

Aborted (core dumped)
```
So for now I've replaced `QApplication` by `QCoreApplication` in pyqt5 tests. However I have no idea why it was working before. I suspect `qapp` variable was somehow shared between pyside2 & pyqt5.

@dbrattli  Also, if you agree, I've replaced `make_app` function by a pytest fixture, and remove dependency to `unittest` (fixtures are not compatible with `unittest.TestCase` subclass anyway).

erikkemperman:
Huh! Yes it should be core application for both... I am pretty sure I changed it in both places locally, maybe I did not commit one of them?

dbrattli:
@erikkemperman I think we should disable Travis build on "Python 3.7 on MacOS, with optional dependencies and benchmark" for now since it's making too much noise.

erikkemperman:
I'll probably have a little time today and I will either address the timings separately, or if I can't get it to work quickly I'll disable the MacOS build as you propose!

jcafhe:
@erikkemperman  For the record, it seems crazy but I confirm that before this PR, the Qt application was shared between pyqt5 & pyside2 tests. So if by chance, pyside2 tests were collected by pytest and run before pyqt5 tests (which was true, before renaming 'qt5' by 'pyqt5', 'pyside2' < 'qt5' regarding alphabetic order),  pyqt5 was getting a QCoreApplication instead of a QApplication:

- pyside2 tests request the current application instance and get None
- instanciate a QCoreApplication and keep a ref at module level
- run pyside2 tests .......
- the application is still alive at level module, and pytest seems to not 'unload' the pyside2 tests module
- pyqt5 tests request the current application instance and get a pyqt5 QCoreApplication wrapping the C++ Application created by pyside2 tests
- run pyqt5 tests .... No problem since it's a QCoreApplication instead of a QApplication

I assume the pytest fixture solve fully this issue by encapsulating the qt application, and avoids imports side-effect.

Sorry for the wall of text.

erikkemperman:
@jcafhe Thanks for sharing! That's actually really helpful. I'm having strange problems with my test refactoring, where the gtkscheduler, wxscheduler and tkinterscheduler tests work well when run in isolation, but as soon as I have more than a single one in the same command strange things happen (crashes or hangs). It might be some similar issue, where artifacts "hang around" in between tests. I'll try if using fixtures (which is probably cleaner anyway) might help!

erikkemperman:
@jcafhe If you rebase this branch onto current master, Travis should pass.

jcafhe:
Thanks for the tips, I will do that.

erikkemperman:
@jcafhe Hmm, I find I am no longer able to run / debug the unittests in Pycharm with this change -- or at least not the way I was used to. I think it is due to the test classes no longer having `unittest.TestCase` as base class. Did you notice something like this, or is it just me?

jcafhe:
Sorry I don't use Pycharm and run pytest in cli. 
I know that pytest actually does not require to have tests declared in a class and support plain functions like:
```python
def test_foo():
    ...

def test_bar(afixture):
    ...
```

I'm not sure but I think tests should still work within a class.
It is probable that pycharm use as a default testing tool `unittest` from the standart library instead of `pytest`?

After a quick search on the internet, try to enable `py.test` in `Preferences>Tools>Python Integrated Tools`.

Anyway, I will wait for your feedback before merging.

erikkemperman:
Thanks for the quick response! I also mostly run them in cli, except when something weird is happening and I like to use the integrated debugger in pycharm. Anyway, it turns out that you were right, it was using the `unittest` test runner instead of pytest (which I assumed it would, being in setup.py and the imports).

Now things work as expected, thanks again!

jcafhe:
Glad you got it resolved!

erikkemperman:
Well, yes and no :-)
The pycharm thing is fine now, and I managed to rebase my scheduler test refactor on your latest work. Now the gtkscheduler, tkinterscheduler and wxscheduler suddenly get along fine (so it was an artifact of the same issue you addressed here I guess?)... But in my branch now qt5 and pyside2 tests can't run together for some reason.

I'll keep at it, sooner or later I'll find the magic bullet!

jcafhe:
@erikkemperman if it helps, I've just run pytest against your branch `scheduler-tests`, commit ee1e78138630f3050a71ab1b7aaa90ccef62c06c and pyside2 & pyqt5 seems to work when `pygobject` is not installed. However, when installing `pygobject` I've got error in pyside2 tests only, with multiple messages:

`
QObject::startTimer: Timers can only be used with threads started with QThread
`
I remember that we need to be extra careful with QtScheduler and QTimer because they cannot be start from an other thread that the main Qt loop.

erikkemperman:
Thanks, yes same here (also linux, at work). But if I invoke only te pyside test it works...

erikkemperman:
And another thing I noticed when running both the qt5 and pyside2 test in the same command: `QCoreApplication.instance()` returns something in the latter, before the constructor is invoked in the pyside2 test fixture... If I just run pyside2 test, this returns something. So something is still carrying over from test into the next? Even though we get QtCore from a different package, or so it seems.

BTW please don’t mind the hopeless mess of “tmp” commits, and so on, just work in progress :-)

jcafhe:
> BTW please don’t mind the hopeless mess of “tmp” commits, and so on, just work in progress :-)
Don't worry ;)

Indeed, there's clearly something that stays in the background. Will try to investigate too.

erikkemperman:
Sorry for spam, but just some findings on this issue:

- It's not pygobject / pycairo I think, I uninstalled them and same error. But the error is not consistent, once in a while things pass (at least on my machine) so I guess it might have appeared to coincide with uninstalling pygobject?
- If I get errors, it is sometimes complaining about the wrong thread, like you mentioned, but other times it actually crashes hard. Inspecting the core dump I find that the pyside2 test is segfaulting in `<venv>/.../site-packages/PyQt5/Qt/lib/libQt5Core.so.5`. So that's the *wrong* core, it should have been from `.../site-packages/Pyside2/...`!

So it looks like the QtCore instance is persisting across tests, even with the fixture trick.

I believe this also happens on current master, actually, and that we are currently still giving PyQt5's core to the QtScheduler constructor during the PySide2 tests :-)

This is fine (well, it's incorrect, but it seems to not be causing failures or crashes) in master. But my scheduler-tests branch is using separate QThread instances to monitor progress (detect when scheduled items have run, or are taking too long) and that extra twist is apparently too much for what seems to be effectively a hybrid of various PyQt5 / PySide2 stuff all active simultaneously.

jcafhe:
> It's not pygobject / pycairo I think, I uninstalled them and same error. But the error is not consistent, once in a while things pass (at least on my machine) so I guess it might have appeared to coincide with uninstalling pygobject?

Yes, it's probable, can't test right now though.

>I believe this also happens on current master, actually, and that we are currently still giving PyQt5's core to the QtScheduler constructor during the PySide2 tests :-)

I'm not sure because if you run `pytest -s` (no capture) against current master and add a print statement in `app` fixture to display the current instance of `QCoreApplication` at initialization time, it returns `None` for pyside2 & pyqt5 tests:
```python
@pytest.fixture(scope="module")
def app():
    print('previous QtCoreApplication instance: {}'.format(QtCore.QCoreApplication.instance()))
    app = QtCore.QCoreApplication([])
    yield app

```

However, with your branch, I've got `None` then `QCoreApplication Object ...`, no matter which of pyqt5 or pyside2 is executed first (although only pyqt5>pyside2 combination crashs).
By the way, I think we should have this kind of warning to detect a possible conflict by requesting `QCoreApplication.instance()` and check if it's `None` before continuing the current test module.

Not sure about that but maybe there's a circular dependency with app<->QThread, need to test if forcing garbage collection could solve this (plus wrapping `wait` in a fixture too). I should have time this evening to test this.

erikkemperman:
Ah! Thank you @jcafhe, you've saved my sanity (for the moment, anyway)!

Yes, I guess the QThread instance was keeping the CoreApplication instance alive, because it has to have it as its parent...

If I make a fixture for the `wait` function, with a dependency injected `app` argument, then I can delete the `app` instance at precisely the right time (as soon as we're done with the QThread).

It looks like this works:
https://travis-ci.org/erikkemperman/RxPY/builds/536322812

Unfortunately I will not be at a computer until late this evening, but thanks again for thinking along with me!

By the way, now that you've seen what I am working towards for the scheduler tests, did you think it might be going in the right direction? I kind of like it that pretty much the same pattern works for a wide variety of (mainloop) schedulers.

Now that this quirk seems to be out of the way, it should not take me too long to do the core schedulers in a similar pattern (that should be a lot easier, which is why I started with the mainloop package). Then I can finally start cleaning things up, making proper commits and submitting a PR :-)

jcafhe:
Yeah that's great! :tada:  I've just tested on my machine too and it works :+1: 

So we should not use (as much as possible) variables at module scope and use fixture instead. By the way,  the [fixture documentation](https://docs.pytest.org/en/latest/fixture.html) is quite a bit long but it's worth reading it IMO. It seems you can achieve a lot of stuff like nested fixtures with multiple level scopes ('session', 'module', ...), or declare global fixture in a conf file which could be initialized according to some environment variables (I was thinking about the `SchedulerTestSetup`).

> By the way, now that you've seen what I am working towards for the scheduler tests, did you think it might be going in the right direction? I kind of like it that pretty much the same pattern works for a wide variety of (mainloop) schedulers.

From what I have understood, it seems to be a great improvement, especially tracking scheduler calls thru `SchedulerHistory`. Although I must say that I quite don't understand the `wait`/`event` mechanism.

erikkemperman:
I will write up a more detailed description when I prepare the PR, but briefly the idea would be that we `wait` from outside of the scheduled callbacks for either completion (the `event` is set from within the callback) or timeout (a failure, the callback is not invoked or something hangs).

And then all the assertions are done in the main thread so there is no chance of 3rd party code eating exceptions.

jcafhe:
Thanks for your insight. That sounds great.

PR #386, opened 2019-05-19
Fix silent on_error when observer throws

dbrattli:
Let exeptions bubble up and mute when on_error/on_completed on the way down.


jcafhe:
Cool!

dbrattli:
@MainRo I was reading the v1 code today and wondered how did this every work? I thought it was the v3 refactoring, but you are probably right that it was a bug in v1 as well.

PR #387, opened 2019-05-20
Align docs with current example code

dbrattli:
Aligning the docs with the current example code. We can discuss how things should be but we should try to have a consistent style in both docs and examples.


dbrattli:
Thanks for reviewing. Let's keep iterating on the docs ...

PR #393, opened 2019-05-23
spilt window operator and add type annotations

jcafhe:
## Update 
This PR split `window` operator in 3 distinct operators, following rxjs naming convention, and use `nonlocal` statement instead of a list:
- `window`
- `window_when`
- `window_toggle`

Also `buffer` operator has been temporarily fix to work with the new `window` operators.
Not sure about that but coverage decrease is probably due to the addition of 3 operators and one deletion, while number of tests remains the same.

## original message
This PR adds type annotation to `window` operator. 

Currently, the window operator supports 3 different signatures as described in [ReactiveX](http://reactivex.io/documentation/operators/window.html) (sorry, I can't link to RxPY specification). Following what @dbrattli  did with `pipe`, `window` is now declared with 3 overloads. However, I'm not an expert with this decorator, so it may be incorrect. I've tested with small examples and it works quite well with `mypy` (Didn't test with `pyre` though). Also I'm not sure if the last redefinition needs to be typed.

@MainRo I've added 2 more marble diagrams and updated the previous one to be consistent with overloads, maybe it take too much space in documentation? Here are what they look like:

close
open
open_close


jcafhe:
Thanks for reviewing. I was wondering if the coverage decrease was a problem? I've checked `window` tests and everything seems to be already tested.

MainRo:
Thanks for the update @jcafhe. Here are few comments but since these operators have a complex behavior, writing simple marble diagrams for them is not easy anyway. Correct me if I am wrong:
 
1st diagram is observable_window_with_closing_mapper. Windows are created on closing observable completion, not on item emission. So arrows alignment should be on completion, not on emission of "c" items. Also since the "c" observables are created from a factory, they may be described as a single labeled observable (However I am not sure this would help the readability of the diagram, so keeping this representation may be better).

2nd diagram is observable_window_with_boundaries. The operator creates an initial window observable before item "a" is emitted (first window is created on subscription to the source observable if I am correct, not when the first item is emitted).

I need to re-read the implementation carefully for the 3rd diagram (observable_window_with_openings). My head is turning upside down with the group_join composition :)

By the way, why don't we expose three different operators instead of implementing them with the same name ? This makes the behavior very hard to explain, we even swap parameters internally (window-openings and window_closing_mapper) depending on their values. Splitting it in several operators would ease the implementation, the typing hints, and the documentation.

I should have some time to work again on documentation next week (I will try to finish the doc pending tasks before the v3 first stable release). I can update these diagrams and add more comments if needed.

dbrattli:
Good suggestion from @MainRo about splitting this operator in 3 and remove the dispatching. This code is probably ported from C# (to JS to Python) at some point where they have extension methods where it's easy to have members with the same name and different signature.

jcafhe:
Ok for splitting, that should greatly simplify the code :+1: 

Concerning operator names, should we just keep names:
`window_with_openings`
`window_with_closing_mapper`
`window_with_boundaries` 

I believe `window_with_openings` should technically be `window_with_openings_and_closing_mapper` but it's a bit long. Or remove "mapper", i.e. `window_with_openings_and_closing`, as well as `window_with_closing`. What do you think?

@MainRo, concerning operator behaviours:
> 1st diagram is observable_window_with_closing_mapper. Windows are created on closing observable completion, not on item emission.

Ok, I thought that the window length and the creation of the next window was bound to the first item emmitted, due to the `take(1)` operator:
https://github.com/ReactiveX/RxPY/blob/442bada3bdd3ef34bb6bece36d4378c6c706dead/rx/core/operators/window.py#L129
So if I understand, we capture all items until one item is emitted by the closing observable, and wait the closing observable completion to start the next window? 

> Also since the "c" observables are created from a factory, they may be described as a single labeled observable

I agree, it's not optimal but I found that having these observables drawn on different lines and horizontally shifted was a good compromise. 

> 2nd diagram is observable_window_with_boundaries. The operator creates an initial window observable before item "a" is emitted

Good catch, will fix it :+1: 

For the 3rd diagram, I wanted to start the resulting observable with a 'node' (as well as starting the boundaries observable with 'a' and not ticks), but there was a graphic bug. I should create a more detailled issue on dooble if you agree.

dbrattli:
Perhaps align with RxJS, e.g:
```
window
window_toggle
window_when
```

https://www.learnrxjs.io/operators/transformation/window.html
https://www.learnrxjs.io/operators/transformation/windowtoggle.html
https://www.learnrxjs.io/operators/transformation/windowwhen.html

jcafhe:
That is great!

jcafhe:
Thanks! I think it should be ok now. 

Sorry for the big diff, but it is basically just about re-indenting the code (have to wrap the old functions), and using non-local statement instead of a list.

Also I've shorten function parameters, i.e:
- `window_boundaries` -> `boundaries`
- `window_closing_mapper` -> `closing_mapper` 

A quick fix in `buffer` is added too, since it is based on the 'old' `window` polymorphic behavior. I guess we should split `buffer` operator too (in an other PR).

MainRo:
>  So if I understand, we capture all items until one item is emitted by the closing observable, and wait the closing observable completion to start the next window?

Yes, as you can see, the new window is created in the on_completed callback.

PR #394, opened 2019-05-24
feat(take_while): add `inclusive` option.

Drakirus:
Thanks for your work. :+1: 
Here a PR that adds an `inclusive` option to the `take_while` operator.
When `inclusive` is set to `True`, the operator emit the value that caused the `predicate` function to return `False`.

### DIFF
--- a/rx/operators/__init__.py
+++ b/rx/operators/__init__.py
@@ -2846,7 +2846,7 @@ def take_until_with_time(end_time: typing.AbsoluteOrRelativeTime,
     return _take_until_with_time(end_time, scheduler=scheduler)
 
 
-def take_while(predicate: Predicate, inclusive: Optional[bool] = False) -> Callable[[Observable], Observable]:
+def take_while(predicate: Predicate, inclusive: bool = False) -> Callable[[Observable], Observable]:
     """Returns elements from an observable sequence as long as a
     specified condition is true. The element's index is used in the
     logic of the predicate function.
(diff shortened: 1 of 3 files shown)


dbrattli:
This is quite a simple fix and it's not easy to compose this functionality from existing operators. But at the same time we should not add every operator possible. Especially when it's easy to make custom operators these days with `pipe`, and this was the reason for adding `pipe`. So the question is if this functionality is so common that it's should be added to RxPY, or if it should be made available from a separate package, or as a blog post telling people who to create such an operator?

https://github.com/ReactiveX/RxJava/issues/1649
https://github.com/monix/monix/issues/871
https://github.com/ReactiveX/rxjs/issues/2420
https://github.com/ReactiveX/rxjs/issues/4000
https://github.com/ReactiveX/rxjs/pull/4115

Drakirus:
I think this functionality is so common that it's should be added to RxPY.

As you said, composing this functionality from existing operators is not a easy task, and it also damage the code readability. 

I'm new to ReactiveX, thus I still have trouble finding the right pipeline/operators to use. Having this functionality available in the main package allows newcomers to easily find the operator they need because everything is centralized in the same documentation page. 

Having a separate package that provides other operators isn't a bad idea. But it means that people will have to look at 2 places for their operators. Having another package will make testing harder..

jcafhe:
I totally agree with @dbrattli in that we should not add new operators which could be implemented by composing existing ones. However, given the links you've provided, this feature seems to be often requested and has been adopted in RxJS and Monix. So I guess it may be worth it.

rx/operators/__init__.py
```diff
@@ -2846,7 +2846,7 @@ def take_until_with_time(end_time: typing.AbsoluteOrRelativeTime,
     return _take_until_with_time(end_time, scheduler=scheduler)
 
 
-def take_while(predicate: Predicate) -> Callable[[Observable], Observable]:
+def take_while(predicate: Predicate, inclusive: Optional[bool] = False) -> Callable[[Observable], Observable]:
```
jcafhe: I think `inclusive` should not be marked as `Optional` since it can't be `None` (well, it can due to the `if` statement but I believe that was not intended?).

Drakirus: fixed!

jcafhe:
@dbrattli Sorry I've just merged this just before you assigned it to yourself. Very sorry

PR #396, opened 2019-05-27
addition of links to community projects supporting backpressure

daliclass:
Hey - There is some good working going on to support backpressure with rxpy 3 I think its a good idea to list the projects in the migration file to support people moving over from 1.6. 

Thoughts?

### DIFF
--- a/docs/additional_readings.rst
+++ b/docs/additional_readings.rst
@@ -1,3 +1,5 @@
+.. _additional_readings:
+
 Additional Readings
 ===================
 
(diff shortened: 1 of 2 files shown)


MainRo:
Could this be moved to the "Additional Readings" sections ? Such libraries are relevant not only when migrating code. The "migration" section can reference the "Additional Readings" section.

daliclass:
@MainRo I like the suggestion, will do this tomorrow 👍

daliclass:
@MainRo finally found the time to get this done, any feedback let me know. Ran it locally with sphinx to validate the links worked as expected

MainRo:
Thank you for the contribution.

PR #399, opened 2019-05-29
Split threadsafe asyncio scheduler in own class

erikkemperman:
Since the `threadsafe` argument is specified at instantiation time, and remains fixed throughout the scheduler's lifetime, it seems to me that we could split this case into its own class, so we won't have to test for this property and route accordingly every time a `schedule` method is called.

Just a minor suggestion, if you prefer the old way that's fine with me too.


jcafhe:
I agree. Maybe in the future (not YAGNI at all), we will be tempted to extend non-theadsafe schedulers with a threadsafe version that requires a completely different implementation, so it makes sense from a contributor point of view.

erikkemperman:
@jcafhe If you agree with these changes, could you approve? It would be easier for some other effort if I could rebase it onto this. Of course if you have reservations or suggestions for improvement, please let me know. Thanks!

erikkemperman:
Thanks for review!

Issue #375, opened 2019-05-11
RxPY v3 Release It!


dbrattli:
The release of v3 is set to June 1, and I would really like to get it released. What are the remaining issues or blockers? 

I don't think everything should be perfect for the release. After the release we should continue to improve RxPY and target a release v3.1 etc as soon as we get feedback on the v3 release.

Lets get it out!

erikkemperman:
I am (still / again) working through the schedulers and their tests, I’m finding some issues which in my opinion should be resolved before 3.0...

Hoping to submit some things soon, but would love your input on #370?

MainRo:
There are some things I wanted to work on on the documentation and the examples, but nothing that would prevent a release. I will try to work on them before the release if I have some time the next weeks.

dbrattli:
Released 3.0.0b1, https://pypi.org/project/Rx/3.0.0b1/

dbrattli:
Beta 2 (3.0.0b2) is at https://pypi.org/project/Rx/3.0.0b2/

jcafhe:
Based on @MainRo proposal to split `window` in 3 distinct operators to simplify type hints and avoid internal dispatching machinery, I suggest that we change few operators: 

[`buffer`](https://github.com/ReactiveX/RxPY/blob/d3ff1b72963fd08341807986d49480351015165e/rx/core/operators/buffer.py#L7) is basically `window` chained with `to_iterable`. So we can split it too with the same naming convention:
- `buffer(openings: Observable)`
- `buffer_when(closing_mapper: Callable[[], Observable])`
- `buffer_toggle(openings: Observable, closing_mapper: Callable[[Any], Observable])`

I believe [`sample`](https://github.com/ReactiveX/RxPY/blob/d3ff1b72963fd08341807986d49480351015165e/rx/core/operators/sample.py#L36) can be simplified too with only 1 argument that accepts an Observable or a time interval. However, a dispatching mechanism will remain but it should be quite simple:
 - `sample(sampler: Union[Observable, RelativeTime)])`

Same with [`catch`,](https://github.com/ReactiveX/RxPY/blob/d3ff1b72963fd08341807986d49480351015165e/rx/core/operators/catch.py#L38) i.e. 1 argument that accepts an Observable or a callable (handler): 
- `catch(handler: Union[Observable, Callable[Exception, Observable], Observable])]`

What do you think? 
I've looked at static operators and it seems fine (except missing type hints, but it won't break API if v3 is released before).

[...]

PR #404, opened 2019-05-30
operators rework: buffer, sample, catch

jcafhe:
As proposed in #375, This PR:
- split `buffer` in 3 distinct operators
- simplify `sample`, the first argument can now be an Observable or a time interval
- simplify `catch`, the 1st argument can be now a handler callable or a n Observable

Also, some `buffer` tests have been added too, since they were missing IMO. I've used marbles testing, and I'm pretty happy, it works ! 

Side note: I'm pretty sure now that for `buffer_when` operator, a buffer is closed and a new one is created as soon as one item is emitted by closing obs, or if the closing obs completes (see tests).  So I guess we can expect the same behavior from `window_when`.

### DIFF
--- a/rx/operators/__init__.py
+++ b/rx/operators/__init__.py
@@ -174,7 +174,7 @@ def buffer_toggle(openings: Observable,
            ---d--|
                        --------e--|
         ----1--2--3--4--5--6--7--8----|
-        [ buffer_toggle()             ]
+        [       buffer_toggle()       ]
         ------1----------------5,6,7--|
 
     >>> res = buffer_toggle(rx.interval(0.5), lambda i: rx.timer(i))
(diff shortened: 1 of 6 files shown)


rx/operators/__init__.py
```diff
@@ -103,31 +103,97 @@ def average(key_mapper: Optional[Mapper] = None) -> Callable[[Observable], Obser
...
+    more buffers.
+
+    .. marble::
+        :alt: buffer_toggle
+
+        ---a-----------b--------------|
+           ---d--|
+                       --------e--|
+        ----1--2--3--4--5--6--7--8----|
+        [ buffer_toggle()             ]
```
erikkemperman: Nitpicking, sorry, but maybe distribute whitespace equally before and after the label inside the marble box?

rx/operators/__init__.py
```diff
@@ -2026,23 +2090,25 @@ def sample(interval: Optional[typing.RelativeTime] = None,
         :alt: sample
 
         ---1-2-3-4------|
-        [     sample(4)    ]
+        [     sample(4) ]
```
erikkemperman: Same here

rx/operators/__init__.py
jcafhe: thanks, will fix it.

dbrattli:
Really nice to get this aligend with `window` operators.

jcafhe:
Thanks for reviewing!

PR #406, opened 2019-05-30
Fix: from_marbles cannot be reused

jcafhe:
When playing with testing marbles, I've noticed a subtle bug with `from_marbles`. Basically, once you subscribe to it and wait for completion, the Observable is 'drained' and cannot be used anymore.

This is due to the fact that the disposable is declared at operator scope, not at subscribe scope. I took the opportunity to do some refactoring too.

### DIFF
--- a/rx/core/observable/marbles.py
+++ b/rx/core/observable/marbles.py
@@ -91,26 +91,30 @@ def hot(string: str, timespan: RelativeTime = 0.1, duetime: AbsoluteOrRelativeTi
     return Observable(subscribe)
 
 
-def from_marbles(string: str, timespan: RelativeTime = 0.1, lookup: Dict = None,
-                 error: Optional[Exception] = None, scheduler: Optional[Scheduler] = None) -> Observable:
+def from_marbles(string: str,
+                 timespan: RelativeTime = 0.1,
+                 lookup: Dict = None,
+                 error: Optional[Exception] = None,
+                 scheduler: Optional[Scheduler] = None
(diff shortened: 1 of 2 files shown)


rx/core/observable/marbles.py
```diff
@@ -91,26 +91,30 @@ def action(scheduler, state=None):
     return Observable(subscribe)
 
 
-def from_marbles(string: str, timespan: RelativeTime = 0.1, lookup: Dict = None,
-                 error: Optional[Exception] = None, scheduler: Optional[Scheduler] = None) -> Observable:
+def from_marbles(string: str,
+                 timespan: RelativeTime = 0.1,
+                 lookup: Dict = None,
```
erikkemperman: Should the Dict have generic type vars (and same in the `parse` function)? I guess we can't do any better than `Dict[Union[int, float, str], Any]` though, so probably not worth it.

jcafhe: Good catch! We can even use `Mapping` instead. I've also missed `Optional` hint for `lookup`.

erikkemperman:
@jcafhe Argh I guess you rebased onto the master branch after I broke it... You could cherry-pick 
f84b596 to make Travis pass?

erikkemperman:
The master branch seems all right again, rebasing on that should make Travis pass.

jcafhe:
Yes I've rebased before seeing your other PR. Will rebase it. Thanks.

PR #407, opened 2019-05-30
Fix silly mistake in online conflict resolve

erikkemperman:
Sigh. I tried github's online merge conflict resolution feature, thinking this was a trivial conflict.

And it was, but I still got it wrong :-(

@jcafhe @dbrattli Apologies, I broke the master branch. This will fix it, but I can't commit it straight to master without approval...

### DIFF
--- a/rx/concurrency/mainloopscheduler/eventletscheduler.py
+++ b/rx/concurrency/mainloopscheduler/eventletscheduler.py
@@ -15,7 +15,7 @@ log = logging.getLogger("Rx")
 eventlet = None
 
 
-class EventLetEventScheduler(PeriodicScheduler):
+class EventletScheduler(PeriodicScheduler):
     """A scheduler that schedules work via the eventlet event loop.
 
     http://eventlet.net/


erikkemperman:
@jcafhe @dbrattli Thanks for the quick response -- I was appropriately distressed whilst the main branch was broken... Apologies, and I'll be more careful.

> Also I was thinking that maybe we should systematically rebase & force push to run tests before merging, because even if there's no conflict in git and code overlapping, there could be conflicts when running tests.

Yes, agreed, and I'll do this regardless of whether the PR was already approved -- I'll never merge anything this way, just because github makes it "convenient", unless local tests as well as Travis pass on the latest commit and with the PR rebased on latest master.

Perhaps it should be a habit to ping those who might have already approved, when PRs change, even if the change appears simple or trivial, and require their re-approval?

jcafhe:
I'm not quite aware of how things are done elsewhere (used to work alone on projects at work) but I guess it happens when multiple people are working on the same project at the same time. On the other side, this is pretty nice to see we have a certain momentum here!

dbrattli:
I'm not to worried about this. We don't have a CI/CD pipeline where changes gets deployed to production, PyPI or anywhere else, so we don't really break anything other than perhaps slowing down the merging of other requests. I like the speed we have, so lets just fail forward and fix the problem when this happens.

Issue #370, opened 2019-05-08
Quick question: schedule_periodic(0.0, ...)


erikkemperman:
Quick question, there seem to be two competing approaches to implementing `schedule_periodic(0.0, ...)`:
- Invoke the action once, and then don't re-schedule
- Invoke the action very often, as quickly as possible

The first is mostly true for the schedulers which inherit the `SchedulerBase` implementation, the second is due to overriding that.

If I want to make things consistent for all schedulers, which approach should be used?

I kind of think the first one is a bit weird, then we are simply doing `schedule()` but with a different callback signature. The second one is probably what people would expect, but might be a bit dangerous (e.g. CurrentThreadScheduler would never return from this call unless one of the action invocations disposes "itself").

jcafhe:
I would say the second is the expected behavior. 

It appears that the current behavior (which I believe is dued to the lines below) was introduced in PR #306, from commit e1fe18a86ef06aab3230486893412ece10d4a307.

https://github.com/ReactiveX/RxPY/blob/93f94ca88da3a7209637b16bb87519b8cc4c396a/rx/concurrency/schedulerbase.py#L129-L130

erikkemperman:
Yes, I think the eventletscheduler did not handle the zero case properly and I had noticed other schedulers doing the first option (one shot).

But I agree, I think it should be the second, if we manage to fix issues that might occur in the various implementations.

dbrattli:
Schedule periodic should schedule as fast as possible (i.e second behaviour). The `schedule_periodic` should be removed from `SchedulerBase` since not all schedulers can or should implement it (e.g CurrentThreadScheduler). In .NET they have a separete interface `ISchedulerPeriodic` used by schedulers that support periodic scheduling.

Thus we should have a separate type e.g SchedulerPeriodicBase with a base impl. of schedule_periodic, and then the schedulers that can impl. periodic scheduling can inherit from this baseclass to signal that they support periodic scheduling, and override if they need to.

PS: Nice read about periodic scheduling (not mentioning the zero case): https://blogs.msdn.microsoft.com/rxteam/2012/06/20/reactive-extensions-v2-0-release-candidate-available-now/

erikkemperman:
All right, thanks, good that there seems to be agreement.

[...]

PR #408, opened 2019-05-31
Rename concurrency package to scheduler, split mainloop and eventloop schedulers

erikkemperman:
As was proposed in https://github.com/ReactiveX/RxPY/issues/370#issuecomment-491516241.

Largish diff again, sorry -- for what it's worth, this was done almost completely by IDE, and then I made sure no lagging references remained by grepping the whole thing for good measure.


erikkemperman:
Thanks!

PR #409, opened 2019-05-31
add type annotations for static operators

jcafhe:
Following operator type annotations #382, it's time to update static operators!

Some details:

Type hint has been relaxed for the `lookup` parameter in  `from_marbles` operator. After a few tests, `Mapping[Union[str, float, int], Any]` seems to not working well or constrains the keys to be of the same type (tested with mypy & pyre). It applies to `case` operator too. 

In the contrary, mixed types seems to work with positional arguments, i.e. `foo(*args: Union[Observable, Future])` will accepts observables mixed with futures. At least, mypy is ok with it. See `on_error_resume_next`.

Some typing.cast directive are added to silence mypy.

### DIFF
--- a/rx/__init__.py
+++ b/rx/__init__.py
@@ -449,8 +449,10 @@ def hot(string, timespan: typing.RelativeTime=0.1, duetime:typing.AbsoluteOrRela
     return _hot(string, timespan, duetime, lookup=lookup, error=error, scheduler=scheduler)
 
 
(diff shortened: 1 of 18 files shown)


jcafhe:
Sorry, I forgot to remove a wrong example in `catch` doc.

jcafhe:
Thanks for reviewing!
Yes, however there are still a lot of errors reported by mypy and a few by pylint . I will try to focus on correcting them.

PR #410, opened 2019-06-01
Inject optional dependencies

erikkemperman:
For consistency among the mainloop/eventloop schedulers, as well as slightly more control over lifetime of references which might clash with one another during tests, I would like to propose adopting a policy of simply requiring the user to pass in modules / instances of the appropriate type into the constructors (usually they would have these available from "outside" anyway).

EDIT Forgot to add, I discovered that what appeared to be “headless” means of testing the `TkInterScheduler` and `WxScheduler` don’t actually work properly — so I’ve changed those to use the regular style (which users would also apply) and enabled “virtual frame buffer” support on Travis to make this work on CI as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,28 +2,45 @@ matrix:
   include:
 
     - name: "Python 3.6 on Linux"
-      python: "3.6"
       os: linux
       dist: xenial
       language: python
+      python: 3.6
+      service:
+        - xvfb
+      script:
+        - xvfb-run -a python3 setup.py test
+
 
     - name: "Python 3.7 on Linux"
-      python: "3.7"
       os: linux
       dist: xenial
       language: python
+      python: 3.7
+      service:
+        - xvfb
+      script:
+        - xvfb-run -a python3 setup.py test
+
 
     - name: "Python 3.8-dev on Linux"
-      python: "3.8-dev"
       os: linux
       dist: xenial
       language: python
+      python: 3.8-dev
+      service:
+        - xvfb
+      script:
+        - xvfb-run -a python3 setup.py test
+
 
     - name: "Python 3.7 on Linux, with optional dependencies and benchmark"
-      python: "3.7"
       os: linux
       dist: xenial
       language: python
+      python: 3.7
+      service:
+        - xvfb
       install:
         # Native libraries for pycairo / pygobject
         - sudo apt-get install -y libgirepository1.0-dev gir1.2-gtk-3.0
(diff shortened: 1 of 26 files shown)


.travis.yml
```diff
@@ -2,28 +2,45 @@ matrix:
...
     - name: "Python 3.6 on Linux"
-      python: "3.6"
       os: linux
       dist: xenial
       language: python
+      python: 3.6
+      service:
+        - xvfb
+      script:
+        - xvfb-run -a python3 setup.py test
```
jcafhe: Noob question here: is the xvfb mandatory here since we don't need a X11 server running (except for jobs with optional dependencies)?

jcafhe:
Nice effort to uniformize things :+1: 
Despite the fact that is a bit less user-friendly, it gives more control/responsibility to end-users and robustness to RxPY. So I agree with this. I would like to hear what @dbrattli  and @MainRo think about this?

.travis.yml
erikkemperman: Yes, unfortunately — the reason is the `TkinterScheduler` tests, which are always run (because tkinter is distributed with python itself).

erikkemperman:
Thanks for review! Yes, this is just a modest proposal, but since would it (slightly) affect API surface I wanted to get it out there now.

Yes I guess it might be a bit less user-friendly, although if they want to use the `GEventScheduler`, for example, I suppose they would have imported `gevent` already anyway and then it’d be simply a matter of adding it to the constructor. And, on the other hand, it gives users more control — for example if there are various implementations available, as we already have with the `QtScheduler`.

I’m also curious what the others think, and if this is deemed “too much” then I completely understand!

.travis.yml
jcafhe: Thanks!

erikkemperman:
Not sure why Travis builds are failing suddenly — except python 3.8, which passes.

erikkemperman:
Well this is annoying, the builds fail on master but I’m actually quite sure that exactly the same commit / history passed on this PR (see above). May be something wrong over at Travis? Unfortunately I have to go out in a little while so I can’t investigate properly :-(

erikkemperman:
@jcafhe @dbrattli 

I’m not at a computer now but was reading Travis logs on the train and wanted to share this quickly — I think the reason of the failures might be a recent pytest release — the (passing) Travis runs on this PR used 4.5 but just today they released 4.6.0  and I notice that the failed Travis run on the merge to master just a while ago was using the new version...

Not sure what the appropriate fix is, but if anyone is affected by this issue I suggest pinning pytest to the 4.5 version for the moment?

PR #411, opened 2019-06-01
Fix mypy errors: Observer, Scheduler, PriorityQueue, SchedulerItem, AsyncSubject

jcafhe:
This PR corrects some of the errors reported by mypy and pylint. A lot of errors remain so I will try to keep PRs as short as possible.

### Observer
Currently, we monkey patch `_on_next_core`, `_on_error_core` and `_on_completed_core` in the `__init__` method. It feels a bit weird. 
If one calls the `Observer` constructor without specifying handler functions, _on_xxx_core methods are ... well, object methods, i.e. `_on_xxx_core(self: object, ...)`.
If handler functions are given in the constructor, _on_xxx_core methods become simple object members which appear to be functions i.e. `on_xxx_core(...)`. 
Since other sub-classes are based on `_on_xxx_core` methods (SchedulerObserver, ObserveOnObserver) like 'protected' pseudo-private methods, I've just added an extra layer of wrapping.
```
rx/core/observer/observer.py:30: error: Name '_on_next_core' already defined on line 19
rx/core/observer/observer.py:44: error: Name '_on_error_core' already defined on line 21
rx/core/observer/observer.py:57: error: Name '_on_completed_core' already defined on line 23
```

Pylint related errors:
```
************* Module rx.core.observer.scheduledobserver
rx/core/observer/scheduledobserver.py:26:4: E0202: An attribute defined in rx.core.observer.observer line 19 hides this method (method-hidden)
rx/core/observer/scheduledobserver.py:31:4: E0202: An attribute defined in rx.core.observer.observer line 21 hides this method (method-hidden)
rx/core/observer/scheduledobserver.py:36:4: E0202: An attribute defined in rx.core.observer.observer line 23 hides this method (method-hidden)
************* Module rx.core.observer.observeonobserver
rx/core/observer/observeonobserver.py:8:4: E0202: An attribute defined in rx.core.observer.observer line 19 hides this method (method-hidden)
rx/core/observer/observeonobserver.py:12:4: E0202: An attribute defined in rx.core.observer.observer line 21 hides this method (method-hidden)
rx/core/observer/observeonobserver.py:16:4: E0202: An attribute defined in rx.core.observer.observer line 23 hides this method (method-hidden)
************* Module rx.core.observer.observer
rx/core/observer/observer.py:30:4: E0202: An attribute defined in rx.core.observer.observer line 19 hides this method (method-hidden)
rx/core/observer/observer.py:44:4: E0202: An attribute defined in rx.core.observer.observer line 21 hides this method (method-hidden)
rx/core/observer/observer.py:57:4: E0202: An attribute defined in rx.core.observer.observer line 23 hides this method (method-hidden)
```

### PriorityQueue
Correct the heap type, i.e. `List[Tuple[T1, int]]`. This can be infer from the `ScheduledItem.enqueue` method:
```python
heapq.heappush(self.items, (item, self.count))
```
```
rx/internal/priorityqueue.py:24: error: Value of type "T1" is not indexable
rx/internal/priorityqueue.py:29: error: Value of type "T1" is not indexable
rx/internal/priorityqueue.py:37: error: Cannot infer type argument 1 of "heappush"
rx/internal/priorityqueue.py:44: error: Value of type "T1" is not indexable
```

### SchedulerItem
It appears that `__eq__` is supposed to accept any type of others, something related with Liskov principle. By the way, I'm sorry for having requesting this strict annotation in PR [362](https://github.com/ReactiveX/RxPY/pull/362#discussion_r279300892).

More readings on mypy issue tracker:
https://github.com/python/mypy/issues/2783
https://github.com/python/mypy/issues/6710
Proposal to add an unsafe decorator:
https://github.com/python/mypy/issues/5704

I think we should keep it as simple as possible, so I changed type hint from `ScheduledItem` to `Any` and implemented approximately the example suggested by mypy.
```
rx/scheduler/scheduleditem.py:43: error: Argument 1 of "__eq__" incompatible with supertype "object"
rx/scheduler/scheduleditem.py:43: note: It is recommended for "__eq__" to work with arbitrary objects, for example:
rx/scheduler/scheduleditem.py:43: note:     def __eq__(self, other: object) -> bool:
rx/scheduler/scheduleditem.py:43: note:         if not isinstance(other, ScheduledItem):
rx/scheduler/scheduleditem.py:43: note:             return NotImplemented
rx/scheduler/scheduleditem.py:43: note:         return <logic to compare two ScheduledItem instances>
```

### missing to_timedelta in rx.core.typing.Scheduler
`to_timedelta` is not declared in rx.core.typing. I took the opportunity to add `to_seconds` and `to_datetime`.
```
rx/subjects/replaysubject.py:43: error: "Scheduler" has no attribute "to_timedelta"
```

### AsyncSubject:
Add some annotations to help mypy. Also, on_dispose, the list of observers is now reset to an empty list. Not sure if it's required but mypy was complaining.
```
rx/subjects/asyncsubject.py:27: error: Need type annotation for 'observers'
rx/subjects/asyncsubject.py:90: error: Incompatible types in assignment (expression has type "Exception", variable has type "None")
rx/subjects/asyncsubject.py:106: error: Incompatible types in assignment (expression has type "None", variable has type "List[Any]")
```

### DIFF
--- a/rx/core/observer/observer.py
+++ b/rx/core/observer/observer.py
@@ -1,6 +1,7 @@
 from typing import Any, Callable, Optional
 
 from .. import typing
+from rx.internal import noop, default_error
 
 
 class Observer(typing.Observer, typing.Disposable):
@@ -9,34 +10,15 @@ class Observer(typing.Observer, typing.Disposable):
     OnCompleted are terminal messages.
     """
 
-    @staticmethod
-    def default_on_next(value: Any) -> None:
-        """Default `on_next()` function. Does nothing."""
-        pass
-
-    @staticmethod
-    def default_on_error(error: Exception) -> None:
-        """Default `on_error()` function. Raises the given exception."""
-        if isinstance(error, BaseException):
-            raise error
-        else:
-            raise Exception(error)
-
-    @staticmethod
-    def default_on_completed() -> None:
-        """Default `on_completed()` function. Does nothing."""
-        pass
-
     def __init__(self,
                  on_next: Optional[typing.OnNext] = None,
                  on_error: Optional[typing.OnError] = None,
                  on_completed: Optional[typing.OnCompleted] = None
                  ) -> None:
-
         self.is_stopped = False
-        self._handler_on_next = on_next or Observer.default_on_next
-        self._handler_on_error = on_error or Observer.default_on_error
-        self._handler_on_completed = on_completed or Observer.default_on_completed
+        self._handler_on_next = on_next or noop
+        self._handler_on_error = on_error or default_error
+        self._handler_on_completed = on_completed or noop
 
     def on_next(self, value: Any) -> None:
         """Notify the observer of a new element in the sequence."""
--- a/rx/core/observer/observer.py
+++ b/rx/core/observer/observer.py
@@ -9,18 +9,34 @@ class Observer(typing.Observer, typing.Disposable):
     OnCompleted are terminal messages.
     """
 
+    @staticmethod
+    def default_on_next(value: Any) -> None:
+        """Default `on_next()` function. Does nothing."""
+        pass
+
+    @staticmethod
+    def default_on_error(error: Exception) -> None:
+        """Default `on_error()` function. Raises the given exception."""
+        if isinstance(error, BaseException):
+            raise error
+        else:
+            raise Exception(error)
+
+    @staticmethod
+    def default_on_completed() -> None:
+        """Default `on_completed()` function. Does nothing."""
+        pass
+
     def __init__(self,
                  on_next: Optional[typing.OnNext] = None,
                  on_error: Optional[typing.OnError] = None,
                  on_completed: Optional[typing.OnCompleted] = None
                  ) -> None:
+
         self.is_stopped = False
-        if on_next is not None:
-            self._on_next_core = on_next
-        if on_error is not None:
-            self._on_error_core = on_error
-        if on_completed is not None:
-            self._on_completed_core = on_completed
+        self._handler_on_next = on_next or Observer.default_on_next
+        self._handler_on_error = on_error or Observer.default_on_error
+        self._handler_on_completed = on_completed or Observer.default_on_completed
 
     def on_next(self, value: Any) -> None:
         """Notify the observer of a new element in the sequence."""
@@ -28,7 +44,10 @@ class Observer(typing.Observer, typing.Disposable):
             self._on_next_core(value)
 
     def _on_next_core(self, value: Any) -> None:
-        pass
+        """For Subclassing purpose. This method is called by `on_next()`
+        method until the observer is stopped.
+        """
+        self._handler_on_next(value)
 
     def on_error(self, error: Exception) -> None:
         """Notify the observer that an exception has occurred.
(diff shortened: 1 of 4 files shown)


rx/core/observer/observer.py
```diff
@@ -9,26 +9,45 @@ class enforces the grammar of observers where OnError and
...
         self.is_stopped = False
-        if on_next is not None:
-            self._on_next_core = on_next
-        if on_error is not None:
-            self._on_error_core = on_error
-        if on_completed is not None:
-            self._on_completed_core = on_completed
+        self._handler_on_next = on_next or Observer.default_on_next
+        self._handler_on_error = on_error or Observer.default_on_error
+        self._handler_on_completed = on_completed or Observer.default_on_completed
```
dbrattli: Here you can use `noop` and `default_error` instead of the static methods. See `AutoDetachObserver`.

dbrattli:
I guess it was monkey-patched to not get the `Observer` into the call stack. Previously I think it was inherited by `AutoDetachObserver` so you would get both `Observer` and `AutoDetachObserver` into the call stack between every operator, so there were some efforts to remove it. But `AutoDetachObserver` does not inherit `Observer` anymore so these changes looks fine. Currently `AutoDetachObserver` is called between every operator so if the oberver blows up we get a call stack like this:

```
  File "/Users/dbrattli/Developer/GitHub/RxPY-release/rx/scheduler/scheduler.py", line 99, in invoke_action
    ret = action(self, state)
  File "/Users/dbrattli/Developer/GitHub/RxPY-release/rx/core/observable/range.py", line 48, in action
    observer.on_next(next(iterator))
  File "/Users/dbrattli/Developer/GitHub/RxPY-release/rx/core/observer/autodetachobserver.py", line 26, in on_next
    self._on_next(value)
  File "/Users/dbrattli/Developer/GitHub/RxPY-release/rx/core/operators/map.py", line 41, in on_next
    obv.on_next(result)
  File "/Users/dbrattli/Developer/GitHub/RxPY-release/rx/core/observer/autodetachobserver.py", line 26, in on_next
    self._on_next(value)
  File "/Users/dbrattli/Developer/GitHub/RxPY-release/rx/core/operators/filter.py", line 34, in on_next
    observer.on_next(value)
  File "/Users/dbrattli/Developer/GitHub/RxPY-release/rx/core/observer/autodetachobserver.py", line 26, in on_next
    self._on_next(value)
  File "test.py", line 9, in error
    raise Exception("Boom!")
Exception: Boom!
```

rx/core/observer/observer.py
jcafhe: thanks! I was looking for something like that but I missed these ones.

jcafhe:
@dbrattli thanks for your insight. Indeed, I've checked and it does not change the stack trace with AutoDetachObserver. 
My first idea was to remove `_on_xxx_core` methods definition in `Observer`, and define them in `__init__` as object members. Thus in the `SchedulerObserver.__init__` constructor, we create local `on_xxx` closure functions by capturing whatever references we need, and pass them to the super constructor i.e. `super().__init__(on_next, on_error, on_completed)`. That would avoid one extra layer of function stack but would clutter `SchedulerObserver`/`ObserveOnObserver` constructors. I guess this is where closure functions and OOP diverge.

@erikkemperman, do you mean in typing.Subject? That should be aligned with `Observable` IMO.
Furthermore, in Observable implementation, `subscribe` accepts 3 handlers between the observer (1st arg) and scheduler (last arg).

erikkemperman:
@jcafhe yes, that’s where I noticed a small divergence with Observable. But maybe it’s slighly more than just the signature, I haven’t checked carefully but perhaps a similar refactor as was done for observer and observable is due for subjects?

erikkemperman:
@jcafhe Did I understand correctly that you were continuing work on this? Let me know if there's anything I can do, particularly the `scheduler` package typing faults are likely due to me :-)

jcafhe:
@erikkemperman No problem. This is not urgent and doesn't block v3 release so I was planning to correct "errors" (which seem to be more wrong typing annotations) little by little. I've just prioritized errors relative to rx/core.

PR #412, opened 2019-06-02
Disallow pytest 4.6.0 for now

erikkemperman:
It appears that pytest 4.6.0 breaks our unittests on Travis (except python 3.8). I propose we disallow that version for the moment.


jcafhe:
Oh sorry, didn't see that you made a PR. Will close #413.
By the way, the issue seems to be identified on pytest tracker:
https://github.com/pytest-dev/pytest/issues/5358

erikkemperman:
Thanks for the quick response, yes I found the same issue.

PR #416, opened 2019-06-03
Subjects refactor

erikkemperman:
It seemed to me that `AnonymousSubject` was not really doing anything -- not exported via the package `__init__.py` and only used from within a unit test. So I propose it be removed.

Further, I thought it was odd that the subject classes inherited from `core.Observable` and `typing.Observer`. I suggest this be changed to `core.Observable` and `core.Observer` as well as `typing.Subject`. I've done this for the main `Subject` class and made the others extend the main class, so structurally this now mirrors what happens in `core.observable` and `core.observer`.

Finally, to align further with the other packages, I would like to suggest we make it `rx.subject` (singular) rather than `rx.subjects` (plural).

Does this make sense?


erikkemperman:
Yes, that was my guess as well — I couldn’t find a usage of this create method other than a test. But if I’m wrong I’ll put it back, obviously!

rx/subject/subject.py
```diff
@@ -30,23 +28,40 @@ def check_disposed(self) -> None:
...
+            if self.exception is not None:
                 observer.on_error(self.exception)
-                return Disposable()
-
-            observer.on_completed()
+            else:
+                observer.on_completed()
             return Disposable()
 
+    def on_next(self, value: Any) -> None:
```
dbrattli: If you inherit from `core.Observer` then you should not impl. `on_next`, but `_on_next_core` since `core.Observer` will then take care fo the stopped logic. Not sure I really like that base-class calling super-class thing, but it was again something inherited from .NET. We should at least keep it similar for all classes inheriting from `core.Observer` for now. Will it work?

erikkemperman: I added a commit which I hope addresses your comment (and I've rebased on latest master).

It looks like this works -- the tests all pass, anyway. But I take your point about calling super-class methods in this case, it does get slightly intricate: The `check_dispose()` calls should happen before looking at `is_stopped`, so the `Subject` class now overrides both `on_next` (to check disposed status) as well as `_on_next_core` (with the actual logic). Likewise, of course, for `on_error` and `on_completed`.

On the other hand, due to the same inheritance one level down, the `XxxSubject` subclasses don't need to know anything about this and need only override the `on_..._core` methods.

rx/subject/replaysubject.py
```diff
@@ -0,0 +1,137 @@
...
+        for observer in observers:
+            cast(ScheduledObserver, observer).ensure_active()
+
+    def _on_error_core(self, error: Exception) -> None:
+        """Notifies all subscribed observers with the exception."""
+
+        with self.lock:
+            observers = self.observers.copy()
+            self.observers.clear()
+            self.is_stopped = True
```
dbrattli: `is_stopped` should already be set by `Observer.on_error`. Do we need to set it here as well? Same with `_on_competed_core`.

erikkemperman: Ah, good catch. No, that's not needed, I've amended the commit.

By the way, I believe the dropped coverage reported by coveralls is simply due to the fact that some code which was at 100% (and remains at 100%) became smaller -- and therefore the "gaps" in the other code become a slightly larger percentage of the whole.

erikkemperman:
Thanks for review, really helpful!

rx/subject/replaysubject.py
erikkemperman: @dbrattli About these spurious dropped coverage reports, I think it looks bad on the list of master commits to have so many red crosses. I'm not not project owner so I cannot do this myself, but apparently you can set the threshold for this on 
https://coveralls.io/github/ReactiveX/RxPY/settings

I would suggest setting this "COVERAGE DECREASE THRESHOLD FOR FAILURE" at something like 0.1%. If that makes sense, could you change it? That seems like a safe margin to assume that anything below that is probably fine.

Note, we'll still see the percentages, I think, it's just that the build status would remain green for percentages less than the threshold.

Thanks!

dbrattli: Thanks. Fixed!

PR #418, opened 2019-06-04
Make sure both class and __init__ docstring are included

erikkemperman:
As mentioned in https://github.com/ReactiveX/RxPY/issues/270#issuecomment-498248593.

By the way, I'm getting warnings about the guzzle theme, it appears they're using an outdated means of injecting script. I've created [an issue](https://github.com/guzzle/guzzle_sphinx_theme/issues/45).

Also, as part of the "final pass" of the docs, I think it would be good to change references to classes and methods using the appropriate markup, this will render them properly and link where possible.

And speaking about links, I will see if I can get the cross-links to work for function parameter types, today. I think that would be a great improvement, but maybe that's just me...

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -60,6 +60,13 @@ extensions = [
     'sphinxcontrib_dooble',
 ]
 
+# Include a separate entry for special methods, like __init__, where provided.
+autodoc_default_options = {
+    'member-order': 'bysource',
+    'special-members': True,
+    'exclude-members': '__dict__,__weakref__'
+}
+
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
 
--- a/rx/subject/behaviorsubject.py
+++ b/rx/subject/behaviorsubject.py
@@ -17,9 +17,9 @@ class BehaviorSubject(Subject):
         creates a subject that caches its last value and starts with the
         specified value.
 
-        Keyword parameters:
-        :param T value: Initial value sent to observers when no other
-            value has been received by the subject yet.
+        Args:
+            value: Initial value sent to observers when no other value has been
+                received by the subject yet.
         """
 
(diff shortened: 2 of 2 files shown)


MainRo:
> Also, as part of the "final pass" of the docs, I think it would be good to change references to classes and methods using the appropriate markup, this will render them properly and link where possible.

Sure. I thought I used the correct syntax (I did not find another way to do it). If this is not correct let's fix it.

erikkemperman:
>  I thought I used the correct syntax

It's something like ```:class:`MyClass` ```I think, although it would sometimes need package prefix as well. I'm not saying it was done incorrectly, by the way, just that I notice some class names and functions which are not yet annotated this way.

I tried again to make the cross-linking work for function arguments and return types, but I guess I simply don't understand why the `sphinx-apidoc`-generated .rst files work and the manual ones that we have currently don't work.

So, I don't know how to resolve this... Well, I'm assuming that we don't want to throw away all the manually organized in favour of a boring tree like [this](https://github.com/erikkemperman/RxPY/tree/apidoc):

Screenshot from 2019-06-04 13-02-35

erikkemperman:
Actually, on second thought... In addition to `__init__`, perhaps *all* dunderscore methods ought to be included, if docstrings are provided?

```python
# Include a separate entry for special methods, like __init__, where provided.
autodoc_default_options = {
    'member-order': 'bysource',
    'special-members': True,
    'exclude-members': '__dict__,__weakref__'
}
```

erikkemperman:
@MainRo Yes I guess I would like all dunderscore methods, when docstrings are provided, should be included by sphinx. I've amended this PR, but since it changed after you approved it, maybe you want to reconsider or reaffirm?

erikkemperman:
Thanks!

PR #419, opened 2019-06-05
Sphinx apidocs: fix crosslink issues, polish rx.Observable

erikkemperman:
So finally, after much annoyance and dead ends, I think I have figured out why the cross-links in the sphinx-generated apidocs were not working
- It obviously only works for classes / methods and so on which we document -- so for instance everything in `rx.core.typing` would never be linked because we weren't refering this module in any rst file.
- There is an issue with re-exporting modules, like is being done for the `Observable` class -- this lives in `rx.Observable`, and is referred to as such in e.g. the docstrings in `rx/__init__.py`, but actually it has a fully qualified name of `rx.core.observable.observable.Observable`. This confuses the sphinx typehints plugin.

The bad news is that this is a bit hackish (only in `docs/conf.py`, *not* in the main code!) but the good news is that if I read this issue correctly some future version will support what I am doing here manually:

https://github.com/agronholm/sphinx-autodoc-typehints/issues/38

I've polished the `Observable` section a bit, and am planning to do the same for `Observable Factory` next.


.bumpversion.cfg
```diff
@@ -1,28 +1,28 @@
 [bumpversion]
 name = Rx
 project = RxPY
-current_version = 3.0.0-beta4
+current_version = 3.0.0b4
```
MainRo: doesn't this break the 'release' scheme ?

dbrattli: I think it needs to match the release values!? Anyways, the when uploaded to pypi it got version v3, even if it was beta3 here. What was the problem?

erikkemperman: Well, setup.py was printing warnings about “normalizing” the version string. And on pypi the value does end up as “3.0.0b4”: https://pypi.org/project/Rx/#history

I actually agree the original version string is nicer, I just have a thing about wanting to get rid of warnings, I suppose.

Also I now realize I haven’t checked if the bumpconfig tool understands this shorter variant, although if it’s apparently the standard way I would assume it does.

Anyway, naturally I’ll change it back if you want!?

erikkemperman: By the way, personally I find I slightly dislike having the project metadata in a file named “bumpversion.cfg” and needing to install a separate tool to edit a value in two places... But there we are.

dbrattli: I agree. We should just remove it. It makes more problems than it solves.

erikkemperman: I've removed the commit about the version string here, let's address it separately (or not).

> We should just remove it.

You mean the tool? And if we remove it, do you think we should have the "full" version string or the shorter one (I think it's not bumpversion which does this normalizing, but setuptools).

dbrattli: Yes, let's remove bumpversion. We have a tool to update a value in two places, and if we remove the tool then we only need to update it in one place? Do we need a tool to do that. I thought it would be simpler with the tool, but it made releasing a new beta so much more complicated.

I'm btw totally agnostic about the format of the version string.

erikkemperman: No, we’d still need to edit in two places. I just think a tool for only that purpose is overkill. I’ll make a proposal when I get to campus.

MainRo: Ok, sorry that this did not meet the goal. On my projects it makes me save time so I thought it would help here. Obviously if this is more a pain than a benefit, let's remove it. The original issue was to avoid reading the config file at runtime, which is still solved with the new proposal of Erik.

MainRo: Ok, sorry that this did not meet the goal. On my projects it makes me save time so I thought it would help here. Obviously if this is more a pain than a benefit, let's remove it. The original issue was to avoid reading the config file at runtime, which is still solved with the new proposal of Erik.

erikkemperman: Actually it did meet the goal, I guess, but as I said my view is perhaps it is a bit much to have a separate tool for editing two strings? And it left important metadata in a file whose name did not convey its purpose (and was hidden). If there’s some benefit to keeping bumpversion that I am overlooking, then please let me know?

erikkemperman:
Thanks for review!

I think the [Observables](https://rxpy.readthedocs.io/en/latest/reference_observable.html) page is in reasonably good shape, but if you have some time please take a look -- the more eyes go over it the less likely it'll be that imperfections fall through the cracks. In particular, it would be helpful if you could be on the lookout for parameter and return value types which are not properly linked internally.

(Just rebased to resolve a conflict, the "copyright" I removed was now in a different file. Otherwise no changes).

PR #420, opened 2019-06-06
Just manage version strings manually

erikkemperman:
As promised in https://github.com/ReactiveX/RxPY/pull/419#discussion_r291033294.

I've added a test to assert that both version strings are the same, as well as comments pointing out that they should remain in sync.

I don't see how this can go wrong :-)


erikkemperman:
Thanks for review!

dbrattli:
I really like the speed here. We don't have to get everything perfect the first time. As long as the PRs brings us in the right direction then we can get some experience and always adjust things later as with this PR. Great work!

PR #424, opened 2019-06-09
Release v3

dbrattli:
I feel it's time to get v3 released. Any last minute changes that need to go in first?

PS: It started with a tweet: https://twitter.com/_mainro_/status/1072965958921277441

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -19,7 +19,7 @@ The Reactive Extensions for Python (RxPY)
 *A library for composing asynchronous and event-based programs using observable collections and
 query operator functions in Python*
 
-RxPY v3.0 Beta
+RxPY v3.0
 ----------------
 
 **This branch is work-in-progress.**
--- a/project.cfg
+++ b/project.cfg
@@ -2,7 +2,7 @@
 name = Rx
 project = RxPY
 # Please make sure the version here remains the same as in rx/__init__.py
-version = 3.0.0b4
+version = 3.0.0
 description = Reactive Extensions (Rx) for Python
 author = Dag Brattli
 author_email = dag@brattli.net
--- a/rx/__init__.py
+++ b/rx/__init__.py
@@ -8,7 +8,7 @@ from .core.typing import Mapper, Predicate
 
 
 # Please make sure the version here remains the same as in project.cfg
-__version__ = '3.0.0b4'
+__version__ = '3.0.0'
 
 
 def amb(*sources: Observable) -> Observable:


erikkemperman:
Still working on some doc polishing and fixing some typehint issues in schedulers. And one bug, albeit a minor one. It would be nice to include them in v3, imho, but I won’t be at a computer until this evening... I understand if you don’t want to wait for it!

erikkemperman:
Oh and the issue about CurrentThreadScheduler is still open... I’m fine either way, but releasing now would mean we break its behavior relative to 1.x — should be at least a conscious decision then, not an accident of release date.

dbrattli:
Ok, we should try to resolve that one first. It's heavy stuf,f but hopefully I will have time to look into it this week. The problem is that I feel a bit agnostic about it, but when in doubt we should align with Rx.NET or RxJava. Let's delay the release until next weekend.

MainRo:
Nothing blocking the release on my side. But I agree that the work on schedulers should be completed before that, to avoid breaking changes just few weeks after the release.

erikkemperman:
The easy way out would be to go back to the old behaviour for `CurrentThreadScheduler` and optionally, if we want to keep it (and use it internally) invent a new class for the new behaviour, e.g. something like `SameThreadScheduler`. The API would be confusing, though.

By the way, sorry for offtopic but it seems to me that there are quite a few branches on the main repo that are stale and possibly could be removed? Would look a bit tidier if we have only 'main' branches here, and keep the working branches on forks?

MainRo:
can we release v3 now that #433 has been merged or should #431 also be merged before the release ?

erikkemperman:
I'm really not sure... #431 does change the API, which is probably not something we should do just after a release -- but I realize it's way too big to review comfortably. I'm sorry about that.

dbrattli:
#431 is a huge PR, but basically it just changes `subscribe` to only take callbacks, the way most of us use subscribe anyways. Thus what we need to decide is if we want this API change or not. The effect is that we really don't need the Observer class anymore, or technically the Observer is really being split into three different type of Observers: OnNext, OnError and OnCompleted e.g:

```
# OnNext observer. Compatible with Callable[[Any], None]
class OnNext(Protocol):
    def __call__(self, value: Any):
        ....
```
I'm still trying to figure out if this is still symmetric to `Iterator`, e.g `Iterable` returns an `Iterator` that has a function `__next__` that can return a value or raise `StopIteration` or some other exception. However to call `__next__()` you should use the builtin `next` where you can provide a default to avoid `StopIteration`.  This is inverted by `Observer` having `on_next`, `on_error` and `on_completed`. However, with Python (and most other languages) you almost never have to deal with iterators directly since they are hidden by the language in constructs such as e.g for-in.

Thus this PR splits the Observer class into three different observers to make it easier to use Callable's as observers. If we go this way we should perhaps remove `Observer` from `abc`
 and replace it with the three smaller callback observers.

MainRo:
So maybe we should postpone this for v4, and take more time to see how to deal with it ? On my side there are still few occasions where I use Observers and not callbacks: do_action debuggers, and subjects are some examples. These cases can be implemented with callback, but the Observer interface simplifies the code. So I have no strong opinion on the way to go.
To get back on v3, I must admit that I wait for an official release before updating my projects (personal and professional ones). Releasing v3 will not be the end of evolutions, but a starting point for more frequent releases, and simpler migration between major releases !

erikkemperman:
Well, as the author of the subscribe refactor I should probably abstain from voting on it :-)

Just to be clear, though, #431 does not remove support for observers, it just gets rid of the polymorphic subscribe implementation and "promotes" the callback style.

I was actually thinking about changing all of the `Callback[...]` typings, such as `ScheduledAction` and so on, to something like you wrote above, just a class with a `__call__` method. I think it would make the documentation a lot more readable, because it would just say `ScheduledAction` (a link to some description of it) instead of a nested mess of square brackets!

But I wasn't sure, wanted to make sure that mypy and IDEs then still understand what's happening...

About releasing: I think the following should at least be considered:

- the docstrings in `operators/__init__.py` have not been polished (e.g., there are examples without `>>>`, functions without argument or return descriptions, typos, ...)
- would be great if @MichaelSchneeberger could verify if issue https://github.com/ReactiveX/RxPY/issues/445 is fixed.

I'm very happy that a beta 5 is now out there, there have been pretty significant changes recently that I would very much like to see at least slightly field-tested before v3 proper.

MichaelSchneeberger:
> would be great if @MichaelSchneeberger could verify if issue #445 is fixed.

done

dbrattli:
Looks like things have stalled. My suggestion is that we release v3 without #431 so we can use more time to get the subscribe handling right and it would not look great if we had to revert it. Any objections?

erikkemperman:
I’m sorry that the subscribe PR is apparently too big to review properly... I mean, I understand — but it was just kind of an all or nothing change, anything less would not have worked at all. It’s a pity, I guess it will have to wait until v4 then, being a breaking change.

erikkemperman:
Congrats on v3 everyone, nice work!

dbrattli:
Congrats to @erikkemperman, @jcafhe, @MainRo and myself. Awesome work, the code is so much better now!! Lets keep master as v3 for a little while to fix issues with the release. Then branch off v3 and use master for v4. I'm not very keen on simultaneously managing a lot of branches so imo all new features should go to v4 and we should aim at releasing it quite soon (this year).

jcafhe:
@dbrattli Thank you for your commitment from the beginning!

MainRo:
great work from all of you. This release is a huge step and RxPY is definitely on par with other implementations ! Hat down to @dbrattli for the courage of dropping the huge work that you did on v2 to directly jump to v3.

PR #425, opened 2019-06-09
Replaced references to rx.scheduler in docs

gargleblaster:
I think there are a many more references to rx.scheduler than these, but I think some are valid and some are not, so I'm afraid to break things.


erikkemperman:
I'm sorry, but the package was renamed from `concurrency` to `scheduler` recently...

dbrattli:
Yes I thought we renamed `concurrency` to `scheduler`.  Why would you rename it back?

PR #426, opened 2019-06-11
Polish docstrings for static operators

erikkemperman:
I notice that the `generate_with_relative_time` and `generate` static ops don't handle a scheduler parameter, but the docstring says:

https://github.com/ReactiveX/RxPY/blob/dc765aaad01cbb4493c280397ff34333cf2158a3/rx/__init__.py#L491-L494

For the moment I've removed the docstring bit about "using the specified" scheduler. But I wonder if an optional `scheduler` argument should be added here?

The newly added marble diagrams for `case` and `for_in` looked a bit "cramped" to my eyes. So I made them slightly wider, but if I missing some particular reason behind this then please let me know (@MainRo)!

Another question, all references to `Observable` in `rx/__init__.py` are actually to `rx.core.observable.Observable` -- probably just because it is already available as that class is being re-exported `rx.Observable`.

But from an API perspective, should the input Observables not actually be `typing.Observable`, so as not to tie the users to our particular concrete base class?

### DIFF
--- a/rx/__init__.py
+++ b/rx/__init__.py
@@ -1,10 +1,10 @@
 # pylint: disable=too-many-lines,redefined-outer-name,redefined-builtin
 
 from asyncio.futures import Future as _Future
-from typing import Iterable, Callable, Any, Optional, Union, Mapping, Sequence
+from typing import Iterable, Callable, Any, Optional, Union, Mapping
 
-from .core import Observable, typing, pipe
-from .core.typing import Mapper, Predicate
+from .core import Observable, pipe, typing
+from .internal.utils import alias
 
 
 # Please make sure the version here remains the same as in project.cfg
@@ -12,7 +12,7 @@ __version__ = '3.0.0b4'
 
 
 def amb(*sources: Observable) -> Observable:
-    """Propagates the observable sequence that reacts first.
+    """Propagates the observable sequence that emits first.
 
     .. marble::
         :alt: amb
@@ -26,10 +26,14 @@ def amb(*sources: Observable) -> Observable:
     Example:
         >>> winner = rx.amb(xs, ys, zs)
 
+    Args:
+        sources: Sequence of observables to monitor for first emission.
+
     Returns:
-        An observable sequence that surfaces any of the given
-        sequences, whichever reacted first.
+        An observable sequence that surfaces any of the given sequences,
+        whichever emitted the first element.
     """
+
     from .core.observable.amb import _amb
     return _amb(*sources)
 
@@ -58,20 +62,21 @@ def case(mapper: Callable[[], Any],
             case statement.
         sources: An object which has keys which correspond to the case
             statement labels.
-        default_source: The observable sequence or Future that will be
-            run if the sources are not matched. If this is not
-            provided, it defaults to rx.Observabe.empty.
+        default_source: [Optional] The observable sequence or Future that will
+            be run if the sources are not matched. If this is not provided,
+            it defaults to :func:`empty`.
 
     Returns:
         An observable sequence which is determined by a case statement.
     """
+
     from .core.observable.case import _case
     return _case(mapper, sources, default_source)
 
 
 def catch(*sources: Observable) -> Observable:
-    """Continues an observable sequence that is terminated by an
-    exception with the next observable sequence.
+    """Continues observable sequences which are terminated with an
+    exception by switching over to the next observable sequence.
 
     .. marble::
         :alt: catch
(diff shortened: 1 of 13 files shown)


rx/__init__.py
```diff
@@ -43,11 +47,11 @@ def case(mapper: Callable[[], Any],
...
 
-        --1---------------|
-        a--1--2--3--4--|
-         b--10-20-30---|
+        ---1---|
+        a----1----2----3----4---|
+         b----10---20---30---|
         [case(mapper, { 1: a, 2: b })]
-        ---1--2--3--4--|
+        -----1----2----3----4---|
```
jcafhe: I think the resulting observable should be shifted to the right (+ 3 hyphens), i.e. the 'a' observable is subscribed when the source observable emits `1`.

rx/__init__.py
```diff
@@ -110,35 +118,43 @@ def catch_with_iterable(sources: Iterable[Observable]) -> Observable:
...
+        subscription function.
 
     .. marble::
         :alt: create
 
         [     create(a)    ]
         ---1---2---3---4---|
 
+    Args:
+        subscribe: [Optional] Subscription function.
```
jcafhe: Should `subscribe` be qualified as `Optional` with `None` as a default value in the signature? Don't know if it's worth to call `create` with no subscribe function thougth.

rx/__init__.py
```diff
@@ -246,47 +275,56 @@ def empty(scheduler: Optional[typing.Scheduler] = None) -> Observable:
...
     .. marble::
         :alt: for_in
 
-        a--1--2-|
-         b--10--20-|
+        a----1----2--|
+         b----10---20--|
         [for_in((a, b), lambda i: i+1)]
-        ---2--3--11--21-|
+        -----2----3----11----21--|
```
jcafhe: Not sure here, but I think there should be 6 hyphens instead of 4 between `3` and `11`, i.e. 2 from the end of 'a' observable + 4 from the start of the 'b' observable.

rx/__init__.py
```diff
@@ -997,9 +1056,8 @@ def using(resource_factory: Callable[[], typing.Disposable],
 
 def with_latest_from(*sources: Observable) -> Observable:
     """Merges the specified observable sequences into one observable
-    sequence by creating a tuple only when the first
-    observable sequence produces an element. The observables can be
-    passed either as separate arguments or as a list.
+    sequence by creating a :class:`tuple` only when the first
+    observable sequence produces an element..
```
jcafhe: Trailing period. Sorry to bother you :trollface:

jcafhe:
Good job! :+1:  

> For the moment I've removed the docstring bit about "using the specified" scheduler. But I wonder if an optional scheduler argument should be added here?

Yes, that could be great, just by adding the common 'pattern' 
`_scheduler = scheduler or scheduler_ or default_scheduler`
 

> But from an API perspective, should the input Observables not actually be typing.Observable, so as not to tie the users to our particular concrete base class?

I agree too.

rx/__init__.py
erikkemperman:
```
        ---1---|
        a----1----2----3----4---|
         b----10---20---30---|
        [case(mapper, { 1: a, 2: b })]
        --------1----2----3----4---|
```

Do you mean like this? I don't know, I tried to keep the vertical alignment the same... I think the point is to emphasize that the result sequence is the same as `a`. But I got this wrong before :-)

rx/__init__.py
erikkemperman:
```
        a----1----2--|
         b----10---20--|
        [for_in((a, b), lambda i: i+1)]
        -----2----3------11----21--|
```
Like this?

rx/__init__.py
erikkemperman: Ah, good catch. Sloppy/paste, I guess... I'll amend.

rx/__init__.py
erikkemperman: No bother, thanks!

erikkemperman:
OK, I've removed the commit about the marble diagrams for now -- I guess I would prefer if @MainRo could take a look, if there's a bit of time, at making them slightly more "balanced" (rather than me risking inadvertently changing something in his effort). It's a really minor point, obviously.

And I think I've addressed the other remarks, about trailing period and incorrect `Optional`.

I'll see if I can quickly propose something about using `typing.Observable` where appropriate, as well as adding `scheduler` arguments where it might make sense.

rx/__init__.py
jcafhe: Exactly :+1:

Issue #430, opened 2019-06-12
Subscribe signature(s)


erikkemperman:
I'm confused about `Observable.subscribe` and `Observable.subscribe_`. If we have the latter, why should the former have to be polymorphic? The reason I ask is I'm trying to reduce the typehint warnings, and mypy is unhappy about the difference between `Subject.subscribe` and the signature in `typing.Observable`...

If I can get things to work, would it make sense to simply have `subscribe` accept a single argument (type `Observer`) and let `subscribe_` handle the case of varying number of callbacks? And actually, the signature in `typing.Observable.subscribe` also seems a bit off, it says the `Observer` argument is optional, but does it make sense to invoke this method without an observer?

I'm probably missing some history here, sorry if this is a silly question...

jcafhe:
This is the rx v1 behavior, i.e. the 2 following examples would work:
`source.subscribe(handler_on_next, handler_on_error, handler_on_complete)`
`source.subscribe(observer)`

This was discussed from this [comment](https://github.com/ReactiveX/RxPY/issues/293#issuecomment-457502444), when callbacks were allowed only in `subscribe_`, and observer in `subscribe`. 
Personally, I prefer `subscribe` with callbacks  (I've never used an Observer as a subscriber). Also I believe @dbrattli attempted to move the code to `subscribe` with callbacks but that would require to update a lot of operators.

erikkemperman:
Ah thanks, I had a feeling this came before but I could not find it!

So my takeaway from re-reading that thread, the consensus seems to be that `subscribe` should take callbacks, and we can have something like `subscribe_observer` to get rid of polymorphism.

Other than the names of the functions, which hopefully my IDE will help me with, I think I’m pretty close to making that work.

dbrattli:
It makes sense to call `subscribe` without observer or callbacks. It's often used when you have side-effects in the chain e.g `flat_map` or `tap` and you just want to start/run things. Note that the name of `subscribe` in the continuation monad is `runCont`. 

Yes, we would need something like `subscribe_observer` to help the operators. The crux is what to do with `abc.Observable`. I would like to avoid 2 subscribe methods. If we only have callbacks then that is ok with me but then the whole purpose of observers goes away (should we eventually remove them?)

mysticfall:
I've started using RxPy recently, and I have a question about the signature of `subscribe` API.

[...]

Issue #293, opened 2019-01-23
Artifacts exposed from `core/__init__.py`


erikkemperman:
Sorry if this is a silly question, but I am slightly confused as to what artifacts are being exposed from `core/__init__.py`. In some cases these are typings (such as Disposable, Scheduler, Observer) and in other cases these are base implementations (e.g. Observable) which  extend the typings.

Would it be better to consistently expose base implementations here, which are presumably more directly relevant for users, and keep the typings only in the package of the same name, for (mostly) internal use?

dbrattli:
Yes, agree. It's a bit messy right now. I'll fix this.

dbrattli:
Added first iteration of fixes. Core still provides ObservableBase as Observable, but we might wan't to reconsider that. What do you think? Other suggestions?

erikkemperman:
Thanks! I guess I would want to rename `ObservableBase` -> `Observable`. If you agree, I can take a stab at that?

dbrattli:
No, wait. I think we should keep `ObservableBase` for now since it's (or should be) the base for `ConnectableObservable`, `GroupedObservable`, `BlockingObservable` and should be for `Observable`, but there's no difference right now. However there's some magic methods in `ObservableBase` that should only be in `Observable`.

erikkemperman:
Not sure I follow completely... I'm going to see if I can figure it out if I look at those classes more closely.

Meanwhile I've been going over core.observer and came up with the following:
https://github.com/erikkemperman/RxPY/commit/53ad49e34b855e2e5c8215e23b587522866eeb7c

But I hadn't read your comment yet -- maybe this commit is not helpful after all?

erikkemperman:
If I understand your comment correctly, you wanted to have those dunderscores / operator overloading in the main exported `Observer` class, but not in the superclass of `ConnectableObservable` and `GroupedObservable` and `BlockingObserable`... Is that correct?

If so, I guess splitting the class would work -- how about something like this?
https://github.com/erikkemperman/RxPY/commit/cede62c2be5ac5fc8e290794953bb47e1daef002

jcafhe:
I've come to the same [conclusion](https://github.com/jcafhe/RxPY3.0-migration-tracking) by listing the methods.

dbrattli:
This is very interresting. But I just arrived at the cabin, so I need to review things tomorrow morning ...

dbrattli:
[...]

PR #431, opened 2019-06-14
Subscribe refactor

erikkemperman:
Ref:
https://github.com/ReactiveX/RxPY/issues/430
https://github.com/ReactiveX/RxPY/issues/293#issuecomment-457502444

I'm a bit conflicted about submitting this -- on the one hand, it seems like too big a patch just before release, on the other hand it would change the API surface and therefore this is probably also something that could not land just after a major release...

But personally I think something like this would improve the API, and if I understand the earlier discussion properly this might be pretty close to consensus to how things ought to be. It seems to me the main reason this wasn't attempted earlier is that it would be quite a tedious thing to do. And so it was :-)

But I got things working, as far as I'm aware, and thus I've decided to just throw this out there, curious what people think.

The idea is basically to adapt the principal user-facing API (`Observable`) to promote the "callback style" of subscribing to an observable. The "observer style" is still supported, and used internally in two ways, but only as implementation details (namely, the operators and the observable sub-class hierarchy).

In other words, the observer style of subscribing only remains in places that could be adapted without touching the observable API.

For what it's worth, the changes here plenty, but they are almost entirely "mechanical":

- Get rid of polymorphism in `Observable.subscribe`, change the signature to only accept an observer. All invokations of `subscribe` that were using callbacks change to `subscribe_`.

- Rename `subscribe` to `subscribe_observer` and rename `subscribe_` to `subscribe`.

- The observer argument to `subscribe_observer` does not have to be optional, if it is not given we may as well invoke the callback-style variant with no arguments. So, considering it compulsory, we can turn things around by moving `Observable.subscribe_observer(observer)` into `Observer.subscribe_to(observable)`.

The last step means that the typing / ABC for Observable now only knows about the callback style. To use the observer style, you have to first have an instance of Observer and subscribe via that instance.

I've run the whole thing through Travis to make sure each of these steps resulted in a passing build, as evidenced by the commit history below.

### DIFF
--- a/notebooks/reactivex.io/startup.py
+++ b/notebooks/reactivex.io/startup.py
@@ -6,7 +6,7 @@ import time
 import inspect
 import logging
 from random import randint
-from rx.testing import marbles
+
 
 logging.basicConfig(format="%(threadName)s:%(message)s")
 log = logging.getLogger("Rx")
@@ -88,7 +88,7 @@ class Subscriber:
 def subs(src, **kw):
     # required for e.g. .multicast:
     obs = Subscriber(src, **kw)
-    subscription = obs.subscribe_to(src)
+    subscription = src.subscribe(obs.on_next, obs.on_error, obs.on_completed)
     if kw.pop('return_subscriber', None):
         return subscription, obs
     return subscription
--- a/notebooks/reactivex.io/startup.py
+++ b/notebooks/reactivex.io/startup.py
@@ -88,7 +88,7 @@ class Subscriber:
 def subs(src, **kw):
     # required for e.g. .multicast:
     obs = Subscriber(src, **kw)
-    subscription = src.subscribe(obs)
+    subscription = obs.subscribe_to(src)
     if kw.pop('return_subscriber', None):
         return subscription, obs
     return subscription
--- a/rx/core/observable/observable.py
+++ b/rx/core/observable/observable.py
@@ -50,13 +50,13 @@ class Observable(typing.Observable):
                         observer: typing.Observer,
                         scheduler: Optional[typing.Scheduler] = None
                         ) -> typing.Disposable:
-        if callable(self._subscribe):
+        if self._subscribe is not None:
             return self._subscribe(observer.on_next,
                                    observer.on_error,
                                    observer.on_completed,
                                    scheduler)
 
-        if callable(self._subscribe_observer):
+        if self._subscribe_observer is not None:
             return self._subscribe_observer(observer, scheduler)
 
         return Disposable()
--- a/rx/core/observable/observable.py
+++ b/rx/core/observable/observable.py
@@ -41,40 +41,25 @@ class Observable(typing.Observable):
                         ) -> typing.Disposable:
         return self._subscribe(observer, scheduler) if self._subscribe else Disposable()
 
-    def subscribe(self,  # pylint: disable=too-many-arguments,arguments-differ
-                  observer: Optional[Union[typing.Observer, typing.OnNext]] = None,
-                  on_error: Optional[typing.OnError] = None,
-                  on_completed: Optional[typing.OnCompleted] = None,
-                  on_next: Optional[typing.OnNext] = None,
+    def subscribe(self,
+                  observer: Optional[typing.Observer] = None,
                   *,
                   scheduler: Optional[typing.Scheduler] = None,
                   ) -> typing.Disposable:
         """Subscribe an observer to the observable sequence.
 
-        You may subscribe using an observer or callbacks, not both; if the first
-        argument is an instance of :class:`Observer <..typing.Observer>` or if
-        it has a (callable) attribute named :code:`on_next`, then any callback
-        arguments will be ignored.
+        If you would like to subscribe using callbacks instead of an observer,
+        please use :func:`subscribe_`.
 
 
         Examples:
             >>> source.subscribe()
             >>> source.subscribe(observer)
             >>> source.subscribe(observer, scheduler=scheduler)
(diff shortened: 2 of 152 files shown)


dbrattli:
Nice work! The thing I'm having problem digesting is the flipped `subscribe_to` that breaks with Rx especially when it needs to be in the abc. Would it perhaps be possible to have it as a plain old function instead (since they all look the same)? Thus we could try eventually get rid of the observers internally and only have callback (but not having to taint the observer interface in the process). E.g having a `subscribe_observer` utility ...

erikkemperman:
Hmm, yes, I guess a standalone utility might work — although simply removing it from the public observer might be sufficient? I think none of this relies on it being in typing.Observer.

I’ll take a look at it!

erikkemperman:
FYI, I have removed the `subscribe_to` from typing and ABC -- so the internal Observer classes still have it, but it's not a requirement for custom implementations.

erikkemperman:
Got rid of a bunch of typing errors. Was hoping that I wouldn't have to revisit all these files again, but can't be helped in this case. Removed `subscribe_to` from core.Observable and added it as static function `internal.util.subscribe(observable, observer)`.

dbrattli:
This is beginning to look really nice! I think it should be possible to also let `_subscribe_core` take callbacks as well? I.e `self._subscribe_core(auto_detach_obs.on_next, auto_detach_obs.on_error, auto_detach_obs.on_completed, , scheduler)`.

erikkemperman:
Yes, that was the intention. But I think maybe not in this PR? It’s already rather huge! My goal here was to get the API-touching part in, the internal bits could be done later and should not need breaking changes... I mean, I’d be happy to try it but I’m worried it would make this even more annoying to review.

dbrattli:
I was thinking about just in `Observable.py`, but I see that we then need to wrap the callbacks back into an observer to avoid the next huge rewrite (and that is where I stopped last time and gave up). So you are right, let's wait with that step.

notebooks/reactivex.io/startup.py
```diff
@@ -88,7 +88,7 @@ def on_completed(self)   : return self._on('cmpl', 'fin')
 def subs(src, **kw):
     # required for e.g. .multicast:
     obs = Subscriber(src, **kw)
-    subscription = src.subscribe(obs)
+    subscription = obs.subscribe_to(src)
```
dbrattli: Revert?

erikkemperman: Ah, yes, that's not right. Thanks!

And actually Subscriber isn't strictly speaking a subclass of `Observer` here, so I changed it to just pass the callbacks from `obs` to `src.subscribe`.

My impressions is the notebooks need to still be aligned to v3, anyway, is that correct?

I've amended, and also found another quirk in delaywithmapper (I think it was broken before, apparently not covered by tests).

rx/core/observable/observable.py
```diff
@@ -23,88 +23,57 @@ class Observable(typing.Observable):
...
                         scheduler: Optional[typing.Scheduler] = None
                         ) -> typing.Disposable:
-        return self._subscribe(observer, scheduler) if self._subscribe else Disposable()
+        if callable(self._subscribe):
+            return self._subscribe(observer.on_next,
+                                   observer.on_error,
+                                   observer.on_completed,
+                                   scheduler)
+
+        if callable(self._subscribe_observer):
```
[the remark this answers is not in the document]
erikkemperman: Ah, yes indeed. Do you think it would be better to check if not None and then just try the call, so an exception will be raised if not callable?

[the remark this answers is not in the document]
erikkemperman: I think you're right, and I've changed it -- now it follows the previous pattern of just checking if the function is set and then attempt to call it. Thanks!!

erikkemperman:
@dbrattli I can see why you gave up on the internal part... I mostly got it to work, but it was pretty annoying. Not really difficult, but *a lot*!

https://github.com/erikkemperman/RxPY/commit/3456a1940802bfaff34e244019f441a6cf6a96eb

I think it's probably better to keep it on the shelf for the moment, though. However, while doing this I ran into a few things that looked off to me and might be legitimate bugs... I'm going to have to leave it here for today, as soon as I have some time I'll discuss those separately.

erikkemperman:
(Rebased onto latest master to resolve merge conflict)

rx/core/observable/observable.py
```diff
@@ -116,11 +85,11 @@ def subscribe_(self,
...
 
         Returns:
-            Disposable object representing an observer's subscription to
-            the observable sequence.
+            Disposable object representing subscription to the observable
+            sequence.
         """
 
-        auto_detach_observer = AutoDetachObserver(on_next, on_error, on_completed)
+        auto_detach_obs = AutoDetachObserver(on_next, on_error, on_completed)
```
dbrattli: Observers are abbreviated as obv. Observables are abbreviated as obs.

erikkemperman:
I rebased this on latest master to resolve merge conflicts. Had to do it twice, did not notice the b5 release. Sorry about all the force-pushes. I also changed the observer abbreviation to `obv` as you mentioned.

For some reason, mypy is raising new warnings now, that were not there on this branch before master, and neither were they there on the master branch. I don't think it's anything particularly scary, but I hope to have some time for a better look at it.

kormang:
Hi!
I also see one more problem with subscribe, PTAL at #565 .

dbrattli:
I'm closing this PR since it's to large and has drifted too far from the current master. I also think the current master fixes some of the same issues ref mypy typing.

PR #440, opened 2019-06-25
Fix mypy errors in rx.scheduler

erikkemperman:
This gets rid of mypy errors for the `rx.scheduler` package. It also includes a minor refactor, namely I thought it might be better to move the temporal converters (`to_datetime` and friends) in a separate place and not couple it to the schedulers so tightly.

### DIFF
--- a/rx/core/abc/scheduler.py
+++ b/rx/core/abc/scheduler.py
@@ -20,3 +20,18 @@ class Scheduler(ABC):
     @abstractmethod
     def schedule_absolute(self, duetime, action, state=None):
         return NotImplemented
+
+    @classmethod
+    @abstractmethod
+    def to_seconds(cls, value):
+        return NotImplemented
+
+    @classmethod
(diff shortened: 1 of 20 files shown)


jcafhe:
Some time ago, I've tried to fix those errors and I've came to almost the same fixs. I like the way you have simplify the `StartableFactory` by removing args.
Same for `PriorityQueue[ScheduledItem[TState]]` in EventLoopScheduler, keeping `Tstate` would require to turn the scheduler to a generic type which seems to be a bit overkill IMO.

erikkemperman:
Thanks for review! Yes I agree, it’s like we need to help mypy more than seems necessary... Partly I think it’s because of `Optional` in the signature arguments, after we check for `None` the type checker still considers that variable to possibly be unset. Perhaps the cleanest way to deal with that is overloading, but that would have been a very tedious change.

PR #441, opened 2019-06-25
Fix mypy errors about overloaded methods

erikkemperman:
It took me a while to figure out what mypy was expecting, precisely, with the overload pattern used in `rx.core.observable.Observable.pipe` and `rx.core.pipe`. Eventually, I found that if the variant of the function which actually implements the logic (the one without the `overload` annotation) has a positional variadic argument `*args` then it must also have a variadic keyword argument `**kwargs`.

Not sure why that is the case, but I have it from one of the authors. Anyway, adding `**kwargs` did made the mypy errors disappear.

But I guess it would be better to somehow prevent those bogus `**kwargs` from showing up in the documentation, so I've made a very simple [sphinx extension](https://github.com/erikkemperman/sphinxcontrib-autodoc-filterparams) which allows to hide designated arguments.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -55,6 +55,7 @@ needs_sphinx = '2.0'
 extensions = [
     'sphinx.ext.autodoc',
     'sphinx.ext.napoleon',
+    'sphinxcontrib_autodoc_filterparams',
     'sphinx_autodoc_typehints',
     'guzzle_sphinx_theme',
     'sphinxcontrib_dooble',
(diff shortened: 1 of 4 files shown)


jcafhe:
It's a bit annoying that we have to modify/add code or directives to correct mypy (& pylint) behavior, and I fear that we will need to fix more and more false errors.

On an other project, I've started to use a mypy conf file `mypy.ini` (`.pylintrc` equivalent) to handle this kind of error "noise", e.g.:

 ```
[mypy]
allow_redefinition = True

[mypy-PyQt5]
ignore_missing_imports = True
```

I'm not saying that we should do this right now, but maybe we should consider ignoring false errors in a `mypy.ini` as much as possible.

**Comment by erikkemperman** [comment]
Agree, although in this case the extension is only cosmetic; I could have just added the `**kwargs` to silence mypy — I just didn’t want to see it in the docs...

**Comment by MainRo** [comment]
With the sphinx extension, the spurious kwargs will not be rendered in the doc, but they will still be displayed by code editors/IDEs (I suppose). So I also think that adding unused parameters and a dedicated sphinx extension to ignore them in the doc is not worth fixing these warnings.

**Comment by erikkemperman** [comment]
Yes they’re still visible if you look at the code, obviously — with a comment that this unused argument is present only to satisfy mypy. I still think it’s worth it to reduce mypy errors as much as possible, though, it does surface real problems here and there, and those become easy to miss or ignore if the output is full of “false positives”.

**Comment by dbrattli** [comment]
The original code had the first overload as a dummy to make pylint think that everything is OK for most cases. This PR removes it, and now pylint starts complaining again. 
Screenshot 2019-06-30 at 11 50 16

However I see that the docstring is now correctly picked up from function itself, so the only problem is that this PR moves the particular overload down so that pylint is not fooled anymore. Note that this overload is not really needed for mypy since `Observable` is also a valid `Any`. 

```
@overload
    # pylint: disable=function-redefined, no-self-use
    def pipe(self,
             *operators: Callable[['Observable'], 'Observable']
             ) -> 'Observable':
        ...  # pylint: disable=pointless-statement
```

If we can move this overload to be the first, then pylint if happy for most cases.

dbrattli:
Typing issues including overloading should be fixed in the latest master. Closing.

PR #442, opened 2019-06-25
Fix various mypy errors

erikkemperman:
This PR fixes miscellaneous mypy errors (I've submitted separately PRs about the scheduler package and the overloading pattern).


erikkemperman:
Thanks for review!

PR #443, opened 2019-06-27
Removed __slots__ from rx.scheduler.currentthreadscheduler._Local.

cdanielw:
This prevented idle and queue from being thread-local.
https://github.com/python/cpython/blob/v3.7.3/Lib/_threading_local.py#L107

### DIFF
--- a/rx/scheduler/currentthreadscheduler.py
+++ b/rx/scheduler/currentthreadscheduler.py
@@ -32,7 +32,6 @@ class Trampoline(object):
 
 
 class _Local(threading.local):
-    __slots__ = 'idle', 'queue'
 
     def __init__(self) -> None:
         super().__init__()


erikkemperman:
Maybe it's just me but wow, that's a bit of a counter-intuitive gotcha... I actually read that same bit of documentation at one time, but found the wording unclear, so I did a quick experiment. I just found it again:

```python3
from threading import current_thread, local, Thread


class MyLocal(local):
    __slots__ = 'name'

    def __init__(self):
        super().__init__()
        self.name = current_thread().name


mylocal = MyLocal()


def test():
    print(current_thread().name, mylocal.name)


if __name__ == '__main__':
    for _ in range(2):
        t = Thread(target=test)
        t.start()
        t.join()
```

This prints two different objects:

```
Thread-1 Thread-1
Thread-2 Thread-2
```

And, for some reason, I was satisfied -- I guess I assumed the `__slots__` entry itself would be shared between threads, but not the values, as they were different.

But you're right, and I was wrong -- if I make each thread sleep and print a couple of times, this becomes apparent:

```python3
from threading import current_thread, local, Thread
from time import sleep


class MyLocal(local):
    __slots__ = 'name'

    def __init__(self):
        super().__init__()
        self.name = current_thread().name


mylocal = MyLocal()


def test():
    for _ in range(2):
        print(current_thread().name, mylocal.name)
        sleep(0.618)


if __name__ == '__main__':
    threads = []
    for _ in range(2):
        threads.append(Thread(target=test))
    for t in threads:
        t.start()
    for t in threads:
        t.join()
```

This prints:
```
Thread-1 Thread-1
Thread-2 Thread-2
Thread-1 Thread-2
Thread-2 Thread-2
```

So my original experiment did not actually ask the right question -- both threads simply see the most recently assigned object...

Thanks for the PR!

PR #444, opened 2019-06-29
Implement flushing support for the distinct operator

grantHarris:
This PR implements flushing support for the distinct operator (similar to how it is implemented in RxJS See: https://rxjs-dev.firebaseapp.com/api/operators/distinct)

### DIFF
--- a/rx/core/operators/distinct.py
+++ b/rx/core/operators/distinct.py
@@ -22,9 +22,13 @@ class HashSet:
             self.set.append(value)
         return ret_value
 
+    def flush(self):
+        self.set = []
+
 
 def _distinct(key_mapper: Optional[Mapper] = None,
-              comparer: Optional[Comparer] = None
+              comparer: Optional[Comparer] = None,
+              flushes: Observable = None,
               ) -> Callable[[Observable], Observable]:
     comparer = comparer or default_comparer
 
@@ -49,6 +53,13 @@ def _distinct(key_mapper: Optional[Mapper] = None,
         def subscribe(observer, scheduler=None):
             hashset = HashSet(comparer)
 
+            if flushes:
+                disposable = flushes.subscribe_(lambda _: hashset.flush(), scheduler=scheduler)
+                source.subscribe_(on_error=lambda _: disposable.dispose(),
+                                  on_completed=lambda: disposable.dispose(),
+                                  scheduler=scheduler
+                                  )
+
             def on_next(x):
                 key = x
 
@@ -60,6 +71,6 @@ def _distinct(key_mapper: Optional[Mapper] = None,
                         return
 
                 hashset.push(key) and observer.on_next(x)
-            return source.subscribe_(on_next, observer.on_error,observer.on_completed, scheduler)
+            return source.subscribe_(on_next, observer.on_error, observer.on_completed, scheduler)
         return Observable(subscribe)
     return distinct
--- a/rx/core/operators/distinct.py
+++ b/rx/core/operators/distinct.py
@@ -1,4 +1,4 @@
-from contextlib import ExitStack
+from contextlib import nullcontext
 
 from typing import Callable, Optional
 from rx.core import Observable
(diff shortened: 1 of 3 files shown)


rx/core/operators/distinct.py
```diff
@@ -22,9 +22,13 @@ def push(self, value):
             self.set.append(value)
         return ret_value
 
+    def flush(self):
+        self.set = []
```
jcafhe: I believe we need some synchronization primitive here, either in `flush`/`push` methods or at caller scope by using the built-in `Observable.lock` (might be tricky).  

Given the following sample, values will be checked or appended to `HashSet.set` on thread A, while the set will be wipe out on thread B. 

```python
import rx

flushes = rx.interval(1.0)

rx.interval(0.4).pipe(
    ops.distinct(flushes=flushes),
    ).subscribe(print)
```

rx/core/operators/distinct.py
```diff
@@ -22,9 +22,13 @@ def push(self, value):
...
+
 
 def _distinct(key_mapper: Optional[Mapper] = None,
-              comparer: Optional[Comparer] = None
+              comparer: Optional[Comparer] = None,
+              flushes: Observable = None,
```
jcafhe: `flushes` should be qualified as `Optional`

rx/operators/__init__.py
```diff
@@ -611,7 +611,8 @@ def delay(duetime: typing.RelativeTime,
 
 
 def distinct(key_mapper: Optional[Mapper] = None,
-             comparer: Optional[Comparer] = None
+             comparer: Optional[Comparer] = None,
+             flushes: Observable = None,
```
jcafhe: should be `Optional` too.

rx/core/operators/distinct.py
grantHarris: Thanks. I'll take a look into that

rx/core/operators/distinct.py
```diff
@@ -47,7 +62,14 @@ def distinct(source: Observable) -> Observable:
         """
 
         def subscribe(observer, scheduler=None):
-            hashset = HashSet(comparer)
+            hashset = HashSet(comparer, lock=flushes.lock if flushes else None)
+
+            if flushes:
+                disposable = flushes.subscribe_(lambda _: hashset.flush(), scheduler=scheduler)
+                source.subscribe_(on_error=lambda _: disposable.dispose(),
```
MainRo: when a *flushes* observable is provided, the source observable will be subscribed twice instead on once. Do we want to dispose the flushes observable on completion/error or on source disposal ? Usually it is done on disposal ( the latter case°. In this case disposal can be modified by adding the disposables of flushes.subscribe and source.subscribe to a a CompositeDisposable, and return this disposable.

grantHarris: Thanks! Updated

rx/core/operators/distinct.py
MainRo: Thanks for the update. I find the usage of ExitStack quite verbose for such a simple usage. using nullconext may be more concise and expressive. For example:

    def flush(self):
        with self.lock or contextlib.nullcontext():
            self.set = []

I let you @jcafhe see what option you prefer.

grantHarris: Ok, I'm cool with that

grantHarris: I agree its a cleaner approach

grantHarris: @MainRo nullcontext seems to be new in 3.7 and breaks the 3.6 build

grantHarris: @MainRo I extracted out nullcontext for 3.6 compatibility. There's almost no code to it, so that might be OK. What do you think?

grantHarris: @jcafhe thoughts?

jcafhe: That looks fine to me, thank you :+1

MainRo: excellent!

MainRo:
yes you are right, The RxJS implementation fails on error from the *flushes* observable:
https://github.com/ReactiveX/rxjs/blob/41888efcd1f34de2ee1eda0bd778d4fe0ab4263e/src/internal/operators/distinct.ts#L86

@grantHarris can you update the PR so that the subscription to *flushes* forwards errors to the observer. should be something like:
flushes_disposable = flushes.subscribe_(lambda _: hashset.flush(), **observer.on_error**, scheduler=scheduler)

grantHarris:
@MainRo Sorry I let this sit for a couple weeks. Vacation and then got busy with some other stuff. I'll make those changes later today 👍

MainRo:
@grantHarris did you have some time for this evolution ?

grantHarris:
@MainRo Sorry about the delay. I got tied up with some other stuff at work and then went on vacation. I'll try to get to it in the next couple days

dbrattli:
Closing as stale PR

PR #446, opened 2019-07-08
 Re-raised thrown exception before passing it to observer.

cdanielw:
This provides the traceback to observable even if the exception was originally raised on a different thread.


cdanielw:
I'm closing this pull request, as the solution provided valid tracebacks in some cases, but failed to do so in other.

PR #447, opened 2019-07-16
Merge scan

cdanielw:
Implemented merge_scan operator. This operator follows the semantics from RxJS [mergeScan](https://rxjs-dev.firebaseapp.com/api/operators/mergeScan). 

I failed to implement this using existing operators. Using scan with an observable seed and an accumulator returning an observable causes the accumulator to be called more often than needed. The implementation is based on existing merge, with some minor tweaks to invoke the accumulator. This caused some code duplication. It's probably possible to come up with some  abstraction solving this, but I didn't want to make more changes than needed in existing code.

Note that I didn't add a max_concurrent option, which RxJS mergeScan does. I fail to understand what such an option would actually do. There are no inner sources to subscribe to, and the accumulator is strictly dependent on the result of previous call.

I created a test-suite with the same tests as for scan. There where some cases I couldn't figure out how to test this properly using the TestScheduler though, so I implemented a TestSubscriber, which subscribes to a regular observable and collects and expose messages. Maybe that was a bit intrusive, but at least it keeps the test-coverage up.

### DIFF
--- a/rx/core/operators/merge_scan.py
+++ b/rx/core/operators/merge_scan.py
@@ -1,0 +1,83 @@
+from typing import Any, Callable
+
+from rx import defer, from_future, of
+from rx.core import Observable
+from rx.core.typing import Accumulator
+from rx.disposable import CompositeDisposable, SingleAssignmentDisposable
+from rx.internal.concurrency import synchronized
+from rx.internal.utils import NotSet, is_future
+
+
+def _merge_scan(accumulator: Accumulator, seed: Any = NotSet) -> Callable[[Observable], Observable]:
+
+    def merge_scan(source: Observable) -> Observable:
+        """Partially applied merge_scan operator.
+
+        Applies an accumulator function, which returns an observable sequence,
+        over an observable sequence and returns each intermediate result.
+
+        Examples:
+            >>> scanned = merge_scan(source)
+
+        Args:
+            source: The observable source to scan.
+
+        Returns:
+            An observable sequence containing the accumulated values.
+        """
+        def subscribe(observer, scheduler=None):
+            accumulator_value = [seed]
+            active = [False]
+            group = CompositeDisposable()
+            is_stopped = [False]
+            queue = []
+
+            def subscribe(xs):
+                subscription = SingleAssignmentDisposable()
+                group.add(subscription)
+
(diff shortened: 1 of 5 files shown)


MainRo:
Thank you for the proposal.
The definition of this operator seems very strange to me: The rxjs documentation states that the accumulator function returns an observable, so the accumulator value should be an observable. However the implementation uses the items emitted by the observables as the accumulator.

I may have missed the aim of this operator, but it seems to me that this cannot work for anything other than trivial use-cases. If the observable returned by the accumulator function does any kind of scheduling or delayed emission, then the value of the accumulator is not deterministic.

On the other way if the accumulator value and the returned observable are separated, then the behavior can be deterministic. This can be implemented simply with a combination of scan and flat_map (e.g. by using a tuple in the scan accumulator value).

Here is an example where the accumulator function returns an observable with a delay (uncomment each "merge" line to see the results):

```
import time
import rx
import rx.operators as ops


def merge_scan(accumulator, seed):
    def _merge_scan(source):
        return source.pipe(
            ops.scan(lambda acc, i: accumulator(acc[0], i), seed=(seed, None)),
            ops.flat_map(lambda i: i[1]),
        )

    return _merge_scan


tick = rx.from_([1, 2, 3, 4])

tick.pipe(
    #ops.merge_scan(lambda acc, i: rx.just(acc+i), seed=0),
    #ops.merge_scan(lambda acc, i: rx.range(acc+i).pipe(
    #    ops.delay(0.1),
    #    ops.do_action(lambda j: print("{}-{}".format(acc+i, j))),
    #), seed=0),

    #merge_scan(lambda acc, i: (acc+i, rx.just(acc+i)), seed=0),
    merge_scan(lambda acc, i: (acc+i, rx.range(acc+i).pipe(
        ops.delay(0.1),
        ops.do_action(lambda j: print("{}-{}".format(acc+i, j)))
    #    ops.map(lambda j: "{}-{}".format(acc+i, j))
    )), seed=0),
).subscribe(
    on_next=print,
    on_error=lambda e: print("Error: {}".format(e)),
    on_completed=lambda: print("Completed!"),
)

time.sleep(3)
```

dbrattli:
Closed because of inactivity

PR #455, opened 2019-08-16
Fixed scan race condition

olivierdeckers:
lock while calling accumulator to prevent race condition in updating the accumulation result

See https://repl.it/repls/QuirkySpottedCommas for an example that shows the bug

### DIFF
--- a/rx/linq/observable/scan.py
+++ b/rx/linq/observable/scan.py
@@ -1,3 +1,5 @@
+from threading import Lock
+
 from rx import Observable
 from rx.internal import extensionmethod
 
@@ -22,6 +24,8 @@ def scan(self, accumulator, seed=None):
     has_seed = seed is not None
 
     source = self
+    
+    lock = Lock()
 
     def defer():
         has_accumulation = [False]
@@ -29,10 +33,12 @@ def scan(self, accumulator, seed=None):
 
         def projection(x):
             if has_accumulation[0]:
-                accumulation[0] = accumulator(accumulation[0], x)
+                with lock:
+                    accumulation[0] = accumulator(accumulation[0], x)
             else:
-                accumulation[0] = accumulator(seed, x) if has_seed else x
-                has_accumulation[0] = True
+                with lock:
+                    accumulation[0] = accumulator(seed, x) if has_seed else x
+                    has_accumulation[0] = True
 
             return accumulation[0]
         return source.map(projection)


jcafhe:
Hi, thanks for your contribution :+1:  

However, I'm not sure that we need synchronization in simple operators like `scan`. Such things should be handled by operators that deals with inner/outer observables.

I guess the issue here comes from `merge_all` (called by `merge` under the hood), since your sample seems to work fine with RxPY v3 (see below). In v3, the locking is done in `merge_all` with the `@synchronized` decorator and a customized [`synchronized`](https://github.com/ReactiveX/RxPY/blob/19b1908724b5a4b7ef3e6c407d52cdc3daaa4ed0/rx/internal/concurrency.py#L10) function:
https://github.com/ReactiveX/RxPY/blob/19b1908724b5a4b7ef3e6c407d52cdc3daaa4ed0/rx/core/operators/merge.py#L79-L125

@dbrattli  @erikkemperman  @MainRo I'm not 100% sure but it would be better to fix `merge_all` instead IMO. 


## Sample v3:
```python
from time import sleep
import rx
from rx.scheduler import NewThreadScheduler

def slow_sum(a, b):
  sleep(0.2)
  return a + b

def fast_sum(a, b):
  return a + b

timer_1 = rx.timer(0.0, 0.010, scheduler=NewThreadScheduler()).pipe(ops.map(lambda x: 1))
timer_2 = rx.timer(0.0, 0.010, scheduler=NewThreadScheduler()).pipe(ops.map(lambda x: 1))

print("When the scan function returns quick enough, we count to 10 correctly")
rx.merge(timer_1, timer_2).pipe(
  ops.scan(fast_sum, 0),
  ops.take(10),
  ops.do_action(lambda x: print(x))
  ).run()


print("When the scan function is slower, it is called with the previous result, therefore we don't count to 10 correctly, but only reach 5")
rx.merge(timer_1, timer_2).pipe(
  ops.scan(slow_sum, 0),
  ops.take(10),
  ops.do_action(lambda x: print(x))
  ).run()
```

MainRo:
I also think that this should be fixed in merge_all since it is the operator that does the actual combination of the streams.

dbrattli:
I cannot repro this anymore so I think it's been fixed by now. I see that `merge_all` is using the `synchronized` wrapper. Anyways feel free to open an issue is this is still a problem.

Issue #467, opened 2019-11-08
Weird interaction between groupBy and to_iterable?


ericbottard:
The following piece of code does not produce what I expect (subscription of numbers whose modulo4 is not zero, because of `.take(3)`):

```python
    def test_broken(self):
        numbers = of(7, 1, 3, 2, 4, 5, 0, 6, 8, 10, 9, 12, 11)
        numbers.pipe(
            op.group_by(lambda x: x % 4),
            op.take(3),
            op.to_iterable(),  
            op.flat_map(lambda list: from_iterable(sorted(list, key = lambda g: g.key))),
            op.do(Observer(on_next=lambda g: print("group modulo=%s %s" % (g.key, g)), on_completed=print("groups complete"), on_error=print_tb)),
            op.flat_map(lambda g: g)
        ).subscribe(on_next=lambda e: print("subscr %s" % e), on_error=print_tb, on_completed=print("THE END"))
        time.sleep(1)
```

Yet, both this
```python
    def test_to_iterable_then_from(self):
        numbers = of(7, 1, 3, 2, 4, 5, 0, 6, 8, 10, 9, 12, 11)
        numbers.pipe(
            op.to_iterable(),
            op.flat_map(lambda list: from_iterable(sorted(list))),
        ).subscribe(on_next=lambda e: print("subscr %s" % e))
        time.sleep(1)

```
which shows that emitting an iterable and then flat_mapping works ok, as well as this
```python
    def test_grouping_behaves_correctly(self):
        numbers = of(7, 1, 3, 2, 4, 5, 0, 6, 8, 10, 9, 12, 11)
        numbers.pipe(
            op.group_by(lambda x: x % 4),
            op.take(3),
            #op.to_iterable(),
            #op.flat_map(lambda list: from_iterable(sorted(list, key = lambda g: g.key))),
            op.do(Observer(on_next=lambda g: print("group modulo=%s %s" % (g.key, g)), on_completed=print("groups complete"), on_error=print_tb)),
            op.flat_map(lambda g: g)
        ).subscribe(on_next=lambda e: print("subscr %s" % e), on_error=print_tb, on_completed=print("THE END"))
        time.sleep(1)
```

which correctly produces
```
groups complete
THE END
group modulo=3 <rx.core.observable.groupedobservable.GroupedObservable object at 0x10a8a3ed0>
subscr 7
group modulo=1 <rx.core.observable.groupedobservable.GroupedObservable object at 0x10a8ab0d0>
[...]

PR #469, opened 2019-11-13
Enhancement: group_by and group_by_until subject mapper

jcafhe:
This PR adds a new parameter `subject_mapper` to operators `group_by` and `group_by_until`.
The purpose of this is to allow the end-user to specify a specialized subject that will be used to push elements for each grouped observable. This factory should be implemented as a function that returns a Subject object.

This is especially useful when chaining `group_by` with `to_iterable`, where a `ReplaySubject` is mandatory instead of a `Subject`. More details in #467 
Example:
```python
rx.of(1, 2, 3, 4, 5, 6).pipe(
    ops.group_by(lambda x: x % 2, subject_mapper=lambda: ReplaySubject())
    ops.to_iterable(),
    ops.flat_map(lambda groups: rx.from_list(groups)), 
    ops.flat_map(lambda group: group),
    ...
```
I've tried to keep theses changes aligned with rxjs6 as much as possible:
https://github.com/ReactiveX/rxjs/blob/master/src/internal/operators/groupBy.ts#L105

More readings from rxjs:
https://github.com/ReactiveX/rxjs/issues/2028
https://github.com/ReactiveX/rxjs/issues/1945

### DIFF
--- a/rx/core/operators/groupbyuntil.py
+++ b/rx/core/operators/groupbyuntil.py
@@ -41,7 +41,7 @@ def _group_by_until(key_mapper: Mapper,
 
     element_mapper = element_mapper or identity
 
-    default_subject_mapper = lambda: Subject()
+    default_subject_mapper = Subject
     subject_mapper = subject_mapper or default_subject_mapper
 
     def group_by_until(source: Observable) -> Observable:
--- a/rx/core/operators/groupbyuntil.py
+++ b/rx/core/operators/groupbyuntil.py
@@ -24,10 +24,13 @@ def _group_by_until(key_mapper: Mapper,
     Examples:
         >>> group_by_until(lambda x: x.id, None, lambda : rx.never())
         >>> group_by_until(lambda x: x.id,lambda x: x.name, lambda grp: rx.never())
+        >>> group_by_until(lambda x: x.id, lambda x: x.name, lambda grp: rx.never(), lambda: ReplaySubject())
 
     Args:
         key_mapper: A function to extract the key for each element.
         duration_mapper: A function to signal the expiration of a group.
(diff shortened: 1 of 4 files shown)


tests/test_observable/test_groupby.py
```diff
@@ -593,6 +593,62 @@ def action2(scheduler, state):
...
+
+        def subscription_even(scheduler, state):
+            source = groups[0]
+            return source.subscribe(observer_even, scheduler=scheduler)
+
+        scheduler.schedule_absolute(500, subscription_odd)
+        scheduler.schedule_absolute(600, subscription_even)
+        scheduler.advance_to(1100)
+
+        assert observer_odd.messages == [
```
MainRo: maybe add a comment: "only the last 2 items of odd/even are received because the ReplaySubject has been configured with a buffer size of 2"

rx/core/operators/groupbyuntil.py
```diff
@@ -37,6 +41,9 @@ def _group_by_until(key_mapper: Mapper,
 
     element_mapper = element_mapper or identity
 
+    default_subject_mapper = lambda: Subject()
+    subject_mapper = subject_mapper or default_subject_mapper
```
MainRo: simply use Subject instead of a lambda ?
```
subject_mapper = subject_mapper or Subject
```

tests/test_observable/test_groupby.py
jcafhe: :+1:  the test is not very self explaining.

rx/core/operators/groupbyuntil.py
jcafhe: thanks :+1:

PR #477, opened 2019-12-15
Fix catch_with_iterable: wrong scheduler used to subscribe to the upstream source

### LINKED ISSUE #476: timeout and retry cause "Do not schedule blocking work!"
I can't figure out how to use **timeout** and **retry** in the same operator chain. I think the timeout is scheduled in "TimeoutScheduler" but the retry moves it back to "Trampolinscheduler" resulting in strange behavior.

# Example

```python
import rx
import rx.operators
import datetime
import time

rx.interval(5.0).pipe(
    rx.operators.timeout(4.0),
    rx.operators.retry()
).subscribe(on_next=lambda x: print("next"), on_error=lambda x: print("Error: " + str(x)))


while True:
    time.sleep(1.0)
```
This example produces unending "Do not schedule blocking work!" log messages originating form TrampolinScheduler.

If I add 
```python
rx.operators.do_action(on_error=lambda x: print("Error: " + str(x))),
```
to the operator chain i get:

> Do not schedule blocking work!
> Error: 'CurrentThreadSchedulerSingleton' object has no attribute 'schedule_periodic'

(Also flooding the log)

I tried using 
```python
scheduler = rx.scheduler.ThreadPoolScheduler(max_workers=10)
````
which works if I use it as scheduler in the subscribe call.

But I'm wondering If this is the correct solution... because I don't really do blocking calls, do I?

I my case I have an Observable that works like a heartbeat and if there is no heartbeat I want to emit a signal.

## Java
* I'm more used to RxJava where this works.

```java
Observable.interval(5000, TimeUnit.MILLISECONDS)
                  .timeout(1000, TimeUnit.MILLISECONDS)
                  .doOnError(er -> System.out.println("Timeout..."))
                  .retry()
                  .subscribe();
```

jcafhe:
This PR fixes `catch_with_iterable` which actually does not forward the downstream scheduler to the upstream observable. This should fix #476.

As a bonus, I've converted the list-hack to `nonlocal` statements. Also I've tried to enforce the naming convention that we had with some other operators, i.e.:
- `scheduler`: scheduler given at declaration time (if any)
- `scheduler_`: scheduler coming from the downstream subscribe
- `_scheduler`: operator scheduler

### DIFF
--- a/rx/core/observable/catch.py
+++ b/rx/core/observable/catch.py
@@ -31,12 +31,13 @@ def _catch_with_iterable(sources: Iterable[Observable]) -> Observable:
 
         subscription = SerialDisposable()
         cancelable = SerialDisposable()
-        last_exception = [None]
-        is_disposed = []
+        last_exception = None
+        is_disposed = False
 
         def action(action1, state=None):
             def on_error(exn):
-                last_exception[0] = exn
+                nonlocal last_exception
+                last_exception = exn
                 cancelable.disposable = _scheduler.schedule(action)
 
             if is_disposed:
@@ -45,8 +46,8 @@ def _catch_with_iterable(sources: Iterable[Observable]) -> Observable:
             try:
                 current = next(sources_)
             except StopIteration:
-                if last_exception[0]:
-                    observer.on_error(last_exception[0])
+                if last_exception:
+                    observer.on_error(last_exception)
                 else:
                     observer.on_completed()
             except Exception as ex:  # pylint: disable=broad-except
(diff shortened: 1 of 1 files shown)


MainRo:
seems good to me. Maybe we could add a unit test on this fix.
Also I wonder how many other places have this bug: We fixed this in several operators, but other ones may also need such a fix.

jcafhe:
> Also I wonder how many other places have this bug: We fixed this in several operators, but other ones may also need such a fix.

Yes I agree and It's a bit scary. However I've inspected few other operators of the same kind and they were ok.

erikkemperman:
I think I've noticed some things which superficially looked like inconsistencies, but did not check in each case if they were using the wrong variable names or actually invoking the wrong scheduler.

Sorry I've been completely inactive for such a long time -- I simply have been way too busy with work/life stuff so no time at all for "hobby" coding...

At one point I had some work done on unifying the scheduler unit tests, and that indeed included testing which schedulers run which actions. It looks like life is settling down a bit finally, so perhaps I'll get around to picking that up where I left off.

Issue #485, opened 2020-01-20
observable is not repeated after concat


shilonosov:
Hi all,

I was trying to port some code I have in c# to python. Especially a piece of code that does error handling and retry. One part of that code is using catch and repeat on observable generated with defer. However, I can't make it work.

It seems to me that **retry** operator is being ignored after **concat** operator.

Here is my code:
```python
from datetime import timedelta, datetime

import rx
from rx import operators as ops
from rx.core.typing import Observer
from rx.scheduler import NewThreadScheduler


class ConsoleObserver(Observer):
    def __init__(self):
        self._start = datetime.now()

    def on_next(self, x):
        print(f"{datetime.now() - self._start} next: {x}")

    def on_error(self, e):
        print(f"{datetime.now() - self._start} error: {e}")

    def on_completed(self):
        print(f"{datetime.now() - self._start} completed")


def main():
    scheduler = NewThreadScheduler()

    ints = rx.interval(period=timedelta(seconds=1), scheduler=scheduler)
    ints = ints.pipe(
        ops.take(3)
    )

    error = rx.of("error?")

    stream = rx.concat(ints, error)
    stream = stream.pipe(
        ops.repeat(2),
    )

    d = stream.subscribe(ConsoleObserver())

    print("Hello world!")
    input()

    d.dispose()


if __name__ == "__main__":
    try:
        main()
    except Exception as e:
        print("got an exception:", e)
```

Actual output:
```
Hello world!
0:00:01.002891 next: 0
0:00:02.002933 next: 1
0:00:03.002997 next: 2
0:00:03.004000 next: error?
0:00:03.005997 completed
```

Expected output:
```
Hello world!
0:00:01.002891 next: 0
0:00:02.002933 next: 1
0:00:03.002997 next: 2
0:00:03.004000 next: error?
0:00:04.002891 next: 0
0:00:05.002933 next: 1
0:00:06.002997 next: 2
0:00:06.004000 next: error?
0:00:06.005997 completed
```

If I replace `stream = rx.concat(ints, error)` with just `stream = ints` I can see the sequence is repeated, which is expected.

Thanks,
Dmitry

jcafhe:
Hi, 
Thanks for your report 👍 . 

Indeed there's a bug in `concat` implementation. The `sources_` iterator is defined outside the `subscribe` scope, so it works only for the 1st subscription. 

[...]

PR #486, opened 2020-01-21
Fix concat:  subscribe scheduler not forwarded  + sources iterable exhaustion 

jcafhe:
This PR fixes the following:
- concat does not forward the subscribe scheduler to the uptream sources. 
- sources iterable is not defined at `subscribe` scope, resulting in iterator exhaustion after the 1st subscription. This is related to #485 

Also this PR enforces the `_scheduler`/`scheduler`/`scheduler_` naming convention.


jcafhe:
closing this for now.

PR #492, opened 2020-02-03
Fix ConnectableObservable: scheduler is not forwarded

### LINKED ISSUE #489: Scheduler provided in subscribe is not used for all observable factories
We now have `Observable.subcribe(..., scheduler=scheduler)`, which is really nice. Let's take the API all the way.

[The documentation says that](https://rxpy.readthedocs.io/en/latest/get_started.html#default-scheduler):

> Operators that accept a scheduler select the scheduler to use in the following way:
>  * If a scheduler is provided for the operator, then use it.
>  * If a default scheduler is provided in subscribe, then use it.
>  * Otherwise use the default scheduler of the operator.

I know that this applies to operators but I think that it would make sense if it also applied to observable factories. This way, we can finally get rid of all the `scheduler=scheduler` boilerplate. See #183 for additional info.

The `timer` observable factory, for instance, still defaults to the `TimeoutScheduler` .

I guess this is a feature request.

jcafhe:
This PR fixes `ConnectableObservable.connect(self, scheduler)` method, where the source observable is not subscribed with the downstream subscribe scheduler.

I guess this is related to the fact that scheduler is a keyword argument only, so scheduler must be given as `subscribe(observer, scheduler=scheduler)`, otherwise it will be dismissed.

A simple test case has been added.

This should fix some operators which are based on `connect(scheduler)` method like `ref_count`, and, by extension, to `share`. This should also fix #489.

### DIFF
--- a/rx/core/observable/connectableobservable.py
+++ b/rx/core/observable/connectableobservable.py
@@ -28,7 +28,7 @@ class ConnectableObservable(Observable):
             def dispose():
                 self.has_subscription = False
 
-            subscription = self.source.subscribe(self.subject, scheduler)
+            subscription = self.source.subscribe(self.subject, scheduler=scheduler)
             self.subscription = CompositeDisposable(subscription, Disposable(dispose))
 
         return self.subscription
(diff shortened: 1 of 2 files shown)


frederikaalund:
Fixes #489

MainRo:
> I guess this is related to the fact that scheduler is a keyword argument only, so scheduler must be given as `subscribe(observer, scheduler=scheduler)`, otherwise it will be dismissed.

Yes, the second argument to the subscribe method is on_error. So scheduler has to be set with kw unless the 3 callbacks are provided before.

Nice!

Issue #480, opened 2020-01-07
(wip) subscribe scheduler forwarding issue


jcafhe:
To continue the discussion started in #477, I've quickly reviewed all the operators in rx/core/operators and I suspect we may have the same issue for the following operators.  ~~Also I've not checked static operators yet, so this list might not be comprehensive~~.

### EDIT:  already fixed operators
- concat_with_iterable: see PR #495  
- join (line 108): see PR #493
- ConnectableObservable: see PR #492 

### delay_with_mapper (line 77)
Subscription delay observable should be subscribed with the same scheduler as the source observable. 
https://github.com/ReactiveX/RxPY/blob/7f6bd8614d2a6a9f109114b29b7b9910fc7b15ae/rx/core/operators/delaywithmapper.py#L77-L80

### ~~expand~~ 
~~don't know exactly, disambiguation needed~~ Seems OK
https://github.com/ReactiveX/RxPY/blob/7f6bd8614d2a6a9f109114b29b7b9910fc7b15ae/rx/core/operators/expand.py#L21-L22

### pairwise (line 40)
https://github.com/ReactiveX/RxPY/blob/7f6bd8614d2a6a9f109114b29b7b9910fc7b15ae/rx/core/operators/pairwise.py#L40

### throttle_first (line 39)
Should be `scheduler_`instead ?
https://github.com/ReactiveX/RxPY/blob/7f6bd8614d2a6a9f109114b29b7b9910fc7b15ae/rx/core/operators/throttlefirst.py#L39

### timeout_with_mapper (line 69)
When the timeout observable completes and 'wins' the race, the other source should be subscribed with the specified scheduler. 
https://github.com/ReactiveX/RxPY/blob/7f6bd8614d2a6a9f109114b29b7b9910fc7b15ae/rx/core/operators/timeoutwithmapper.py#L69

### to_marbles (line 57)
https://github.com/ReactiveX/RxPY/blob/7f6bd8614d2a6a9f109114b29b7b9910fc7b15ae/rx/core/operators/tomarbles.py#L57

PR #493, opened 2020-02-03
Fix join: subscribe scheduler is not forwarded

jcafhe:
this PR attempts to fix the `join` operator to continue the ongoing effort to resolve scheduler forwarding issues (#480). 

Bonus:
- convert list hack to `nonlocal` statement
- add a test case: check that the subscribe scheduler is propagated to left & right observables as well as duration observables


jcafhe:
Thanks for the review. I've just fixed the assertion statement ( `is` instead of `==` for identity assertion) and added a test case (no scheduler in subscribe method).

PR #494, opened 2020-02-03
Fixed documentation build with typehints

MainRo:


MainRo:
The docs build was broken since sphinx-autodoc-typehints 1.8 release because a new parameter has been added to format_annotation.

Issue #505, opened 2020-04-15
Concurrency issue using zip operator


mmpio:
Hello, 
The piece of code below produces an exception "pop from empty list".
```python
import rx.scheduler
import rx.operators
import rx.disposable
import time
import typing


class Generator:
    def __init__(self):
        self.__index = 0
        self.__acquire = True

    def generate(self) -> 'rx.Observable':
        return rx.create(
            self.__generate
        )

    def __generate(self,
                   observer: 'rx.typing.Observer',
                   scheduler: 'typing.Optional[rx.typing.ScheduledAction]') -> 'rx.typing.Disposable':
        while self.__acquire:
            observer.on_next(self.__index)
            self.__index += 1

        return rx.disposable.Disposable(action=self.__dispose)

    def __dispose(self):
        pass


if __name__ == '__main__':
    rx.zip(
        Generator().generate().pipe(rx.operators.subscribe_on(scheduler=rx.scheduler.NewThreadScheduler())),
        Generator().generate().pipe(rx.operators.subscribe_on(scheduler=rx.scheduler.NewThreadScheduler())),
    ).subscribe(
        on_next=lambda x: print(x),
        on_error=lambda e: print(e),
        on_completed=lambda: print("onCompleted")
    )
    time.sleep(3600)

```

I have proposed a fix : #506

PR #506, opened 2020-04-18
Fix race condition in zip operator

mmpio:
I have faced race conditions using the zip operator.
The list which stores the zipped observables values may be accessed concurrently, thus, checking and popping data from that list shall be done in an atomic way.

### DIFF
--- a/rx/core/observable/zip.py
+++ b/rx/core/observable/zip.py
@@ -4,6 +4,9 @@ from rx import from_future
 from rx.core import Observable, typing
 from rx.disposable import CompositeDisposable, SingleAssignmentDisposable
 from rx.internal.utils import is_future
+from rx.internal.concurrency import synchronized
+from threading import RLock
+
 
 # pylint: disable=redefined-builtin
 
@@ -28,9 +31,11 @@ def _zip(*args: Observable) -> Observable:
 
     def subscribe(observer: typing.Observer, scheduler: Optional[typing.Scheduler] = None):
         n = len(sources)
-        queues : List[List] = [[] for _ in range(n)]
+        queues: List[List] = [[] for _ in range(n)]
+        lock = RLock()
         is_done = [False] * n
 
+        @synchronized(lock)
         def next(i):
             if all([len(q) for q in queues]):
                 try:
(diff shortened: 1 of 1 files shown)


mmpio:
@MainRo 
I was wondering that too. But if for any reason, zip is called with no arguments, then we'll face a "tuple index out of range" trying to access args[0].lock.

If you decide to keep the solution I've submitted, the lock may even be created in the subscribe function. I don't know what is the best implementation.

jcafhe:
I believe source locks are just helpers, so I guess you may use a local lock with no issue. 

@mmpio @MainRo  I was wondering if appending value to that same list should be controlled by the same lock too?
https://github.com/ReactiveX/RxPY/blob/82bc96059ab9f4ed0d0c93a15095c175fcff0e5d/rx/core/observable/zip.py#L63-L65

mmpio:
@jcafhe I think it is a good idea to move the 'synchronized' decorator from the 'next' function to the 'on_next' function. As the latter is the only one to call the first, the performance will be the same, but at least we won't need to think about the thread issues that may rise without the lock control on appending.

Speaking about performances, the zip operator also runs way faster with the lock, at least in my sample code #505

jcafhe:
> Speaking about performances, the zip operator also runs way faster with the lock, at least in my sample code #505

Indeed, on my laptop and a sleep time of 1s, I get around 500-800 values without the lock, and ~25000 values with the lock.

Also, I was thinking that we generally use a re-entrant lock, i.e. `threading.Rlock` instead of a `threading.Lock`. I don't know if it' mandatory here, but there may be some corner cases especially with some recursive mechanisms in rx. What do you think?

MainRo:
🤔 how can you have much better performances with an additional lock ? That is strange. What bench do you run ?

Concerning the lock on append, this is an atomic operation (https://docs.python.org/3/faq/library.html#what-kinds-of-global-value-mutation-are-thread-safe), so we could keep the lock only on next.

On using RLock, you are right. This is the one used in Observable. It is useful only on cycle graphs where zip depends on its own output, but everything is possible !

mmpio:
We used the script at #505

PR #509, opened 2020-04-25
Drop reference to cached items asap in to_xxx

MainRo:
This reduces memory consumption on cases where the cached items are
quite big and: The observable is not disposed immediately or the
completion callback triggers other computations.


MainRo:
I have cases using to_list where the resulting list is quite big (several GB). The result is then processed in the completion callback. This change allows to free the list as soon as it is not used by the downstream computations instead of keeping it as long as the program runs.

PR #511, opened 2020-04-26
(wip) fix travis jobs py37 with optional dependencies (linux & windows)

jcafhe:
**Work in progress**

### Linux build 3.7 with dependencies:
- update pip before installing PyQt5 should fix the PyQt5 error, no need to forbid specific version anymore. 
https://github.com/python-pillow/Pillow/pull/4527
- pip try to install wxpython 4.1.0 from extra repo https://extras.wxpython.org/wxPython4/extras/linux/gtk3/ubuntu-16.04, but the last version available which is compatible with python 3.7 is wxpython 4.0.7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ matrix:
         # update pip as a workaround to PyQt5.14 installation failure
         - pip3 install --upgrade pip
         # Pre-built wheel for wxpython
-        - pip3 install -U -f https://extras.wxpython.org/wxPython4/extras/linux/gtk3/ubuntu-16.04 wxpython
+        - pip3 install -U -f "https://extras.wxpython.org/wxPython4/extras/linux/gtk3/ubuntu-16.04" wxpython~=4.0.0
         # Install the various optional packages
         - pip3 install eventlet gevent pycairo pygobject pygame pyqt5 pyside2 tornado twisted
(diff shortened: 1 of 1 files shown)


MainRo:
Should I wait that this PR is merged before merging the pending ones with failing CI ?

jcafhe:
@MainRo  Good question! Right now I think errors related to Qt and Wxpython are fixed. 

However some errors related to twisted still remain, and I know almost nothing about twisted. I've tried to dig a bit but it's too much for me. I don't know if these errors are legitimate.

So I would say that you should not wait for this PR before merging others.

MainRo:
The twisted test pass when pytest-twisted is installed. I guess we just need to add it in the travis config with optional dependencies.

MainRo:
up @jcafhe ? see my previous comment. With this fixed and some of the pending PR, it would be nice to do another bugfix release.

jcafhe:
@MainRo yep sorry for the delay, I'm quite busy right now, but I will try to fix and rebase this tomorrow :+1:

jcafhe:
I still get twisted errors on linux and now there's 1 error on MSW with wxpython. I'm a bit desperate. I think we should disable these tests (wxpython and twisted) or maybe update tests with optional dependencies to py3.8 (didn't try, though).

PR #519, opened 2020-06-10
Avoid GEventScheduler deadlock

p4l1ly:
With the following minimal repro, `gevent==20.6.0`, `Python 3.8.2`, the function `buffer_with_time` did not work correctly:

```
import gevent
import rx
from rx.scheduler.eventloop import GEventScheduler

sched = GEventScheduler(gevent)

rx.interval(0.05).pipe(
    ops.buffer_with_time(0.2, scheduler=sched),
).subscribe(on_next=print)

gevent.sleep(2)
```

The changes in this pull request fix the behaviour.

### DIFF
--- a/rx/scheduler/eventloop/geventscheduler.py
+++ b/rx/scheduler/eventloop/geventscheduler.py
@@ -52,7 +52,7 @@ class GEventScheduler(PeriodicScheduler):
         timer = self._gevent.spawn(interval)
 
         def dispose() -> None:
-            timer.kill()
+            timer.kill(block=False)
 
         return CompositeDisposable(sad, Disposable(dispose))
 
@@ -86,7 +86,7 @@ class GEventScheduler(PeriodicScheduler):
         timer = self._gevent.spawn_later(seconds, interval)
 
         def dispose() -> None:
-            timer.kill()
+            timer.kill(block=False)
 
         return CompositeDisposable(sad, Disposable(dispose))
 


AXGKl:
error pretty much independent of python3/gevent versions, also e.g. with 3.7/1.4

=> hefty error & fix looks correct to me

PS 1: 'did not work correctly' means: stream stops deadlocked, after the first buffered emission.
PS2: Glad to see gevent and Rx in use together ;-)

jcafhe:
thanks for the fix 👍 .

I must say that I'm absolutely not proficient in gevent but after digging a bit in the documentation, I was wondering if replacing `greenlet.kill()` by `gevent.kill(greenlet)`  was some sort of workaround? I mean maybe it's obvious, but could you share some insights on this @p4l1ly  @AXGKl before approving the PR?

From the documentation of gevent.kill http://www.gevent.org/api/gevent.greenlet.html#gevent.Greenlet.kill

>**kill(greenlet, exception=GreenletExit)**[source]
Kill greenlet asynchronously. The current greenlet is not unscheduled.

>Note **The method Greenlet.kill() method does the same and more**(and the same caveats listed there apply here). However, the MAIN greenlet - the one that exists initially - does not have a kill() method, and neither do any created with spawn_raw(), so you have to use this function.
Caution Use care when killing greenlets. If they are not prepared for exceptions, this could result in corrupted state.
Changed in version 1.1a2: If the greenlet has a kill method, calls it. This prevents a greenlet from being switched to for the first time after it’s been killed but not yet executed.
```

**Comment by p4l1ly** [comment]
Actually, gevent is obscure to me too. I don't use it much and I don't like it, I only needed to contribute to some older code. I was hoping that a gevent pro would appear here and explain why this fix seem to work and why it was "deadlocking" before. I tried to simulate the problem (a greenlet that kills itself) with pure gevent and I did not succeed (it did not deadlock). So... I'm afraid that we'll need to postpone this merge request until someone, perhaps me if I find some more time, manages to explain it in detail.

Maybe it would be sufficient to use `Greenlet.kill(blocking=False)`? But it's just a guess for now.

**Comment by jcafhe** [comment]
Thank you for that clarification.
I've done some research on the gevent scheduler and I believe it was added in RX v1 in commit https://github.com/ReactiveX/RxPY/commit/65478a4b483bf51b738b896f3f43a886395d51f7

According to this comment https://github.com/ReactiveX/RxPY/issues/5#issuecomment-53184296 :
>dbrattli commented on 24 Aug 2014
> Experimental support for asyncio, ioloop and gevent now in latest 0.6 release. You can see the impl. https://github.com/dbrattli/RxPY/tree/master/rx/concurrency/mainloopscheduler, and examples on how to use at https://github.com/dbrattli/RxPY/tree/master/tests/test_concurrency/test_mainloopscheduler. Please test so we can fix any errors.

I'm wondering if this scheduler has ever been used extensively by someone before? If not, I guess we can safely merge this PR for someone who really needs it.  What do you think @MainRo @dbrattli @erikkemperman ?

**Comment by MainRo** [comment]
I did not comment before because I also never used gevent :)
However I have few comments from the discussion:

1- Original issue
On the original issue, you say that the code blocks. However you provide the gevent scheduler only to the buffer_with_time operator. So the interval operator will still use the timeout scheduler. Not sure but this may cause cohabitation issues with gevent. Can you try this:

```
sched = GEventScheduler(gevent)

rx.interval(0.05).pipe(
    ops.buffer_with_time(0.2),
).subscribe(
    on_next=print,
    scheduler=sched
)
```

This would force all operators to use gevent. However it is strange that there is no explicit call to start the gevent event loop (I did not find how it is started by reading quickly the docs).

2 greenlet.kill vs gevent.kill
From the documentation both API are similar. However, gevent.kill is asynchronous. So I suspect that replacing "timer.kill" with "gevent.kill(timer)" is similar to using "timer.kill(block=False). Here again depending on how gevent work it may or be the correct solution.

p4l1ly:
@MainRo, to answer your comments:
1. I've tried your snippet and it didn't help to avoid deadlocking. And the gevent loop is started implicitly somewhere in `import gevent`. Citing [this docs page](http://www.gevent.org/intro.html#event-loop): "Unlike other network libraries, though in a similar fashion as eventlet, gevent starts the event loop implicitly in a dedicated greenlet."

2. You're right, I've just tried it with `timer.kill(block=False)` and the deadlock disappeared. I'm going to update my fix because the new version resembles more the original code. And as none of us deeply understands gevent, it should be safer to change as little as possible.

PR #527, opened 2020-07-09
[rx] [core] [observable] [catch] fix coding convention 

JAY-Chan9yu:
Hi! first! Thank you! Because I am using RxPy very useful and fun:D

when i read code in 'core/observable/catch', I found typo in violation of PEP8

There is no space to the left of '=' in this code.

Thank you! Have a good Day!


techinpark:
nice catch 👍

Issue #502, opened 2020-03-25
How to convert observable to cancel-able future


puzzleshark:
In the to_future operation, there is a note that cancellation will not work. How can you create a future from an observable that can be cancelled?

MainRo:
This is not supported now, but we could add a done callback and dispose in this callback.

MainRo:
This has been fixed in #529

PR #529, opened 2020-07-16
fix InvalidStateError in to_future

frederikaalund:
Motivating example:

```python
import asyncio
import rx

async def _main():
    obs = rx.just(42)
    fut = next(obs.__await__())
    fut.cancel()
    await obs

asyncio.run(_main())
```

The above raises the following exception:
```python
Exception in callback None()
handle: <Handle cancelled>
  File "/usr/lib/python3.8/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File "/home/fpa/.local/lib/python3.8/site-packages/rx/scheduler/eventloop/asyncioscheduler.py", line 48, in interval
    sad.disposable = self.invoke_action(action, state=state)
  File "/home/fpa/.local/lib/python3.8/site-packages/rx/scheduler/scheduler.py", line 103, in invoke_action
    ret = action(self, state)
  File "/home/fpa/.local/lib/python3.8/site-packages/rx/core/observable/returnvalue.py", line 31, in action
    observer.on_completed()
  File "/home/fpa/.local/lib/python3.8/site-packages/rx/core/observer/autodetachobserver.py", line 44, in on_completed
    self._on_completed()
  File "/home/fpa/.local/lib/python3.8/site-packages/rx/core/operators/tofuture.py", line 48, in on_completed
    future.set_result(last_value)
asyncio.exceptions.InvalidStateError: invalid state
```

The example is contrived but I encountered the same issue in production.

### DIFF
--- a/rx/core/operators/tofuture.py
+++ b/rx/core/operators/tofuture.py
@@ -38,18 +38,20 @@ def _to_future(future_ctor: Optional[Callable[[], Future]] = None,
             has_value = True
 
         def on_error(err):
-            future.set_exception(err)
+            if not future.cancelled():
+                future.set_exception(err)
 
         def on_completed():
             nonlocal last_value
-            if has_value:
-                future.set_result(last_value)
-            else:
-                future.set_exception(SequenceContainsNoElementsError())
+            if not future.cancelled():
+                if has_value:
+                    future.set_result(last_value)
+                else:
+                    future.set_exception(SequenceContainsNoElementsError())
             last_value = None
 
-        source.subscribe_(on_next, on_error, on_completed, scheduler=scheduler)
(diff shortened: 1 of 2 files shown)


frederikaalund:
I now also dispose the subscription when the future is done. This avoids leaking subscriptions on cancellation. See #502 for reference.

I've removed the comment that said "No cancellation can be done" since cancellation is now supported.

MainRo (changes requested):
Thanks, that seems good to me.
Can you add some unit tests ?

frederikaalund:
Thanks for having a look at this issue.

> Can you add some unit tests ?

I have added two unit tests.

Let me know if you need something else.

MainRo (approved):
Thanks!

PR #550, opened 2021-01-08
fix github actions

MainRo:
Now that github actions is enabled, I will use this PR to fix all jobs.


MainRo:
This step is almost ok:

@dbrattli can you create a github secret variable containing a coveralls API token ? See in coveralls doc here:
https://coveralls-python.readthedocs.io/en/latest/usage/configuration.html#github-actions-gotcha

@jcafhe @erikkemperman I disabled the type hinting tests (f82) for now. I think I will create a dedicated issue to re-enable it later because it seems to be broken from python 3.9.

MainRo:
Finally, things seem to be good: I ran the Ci 10 times without failure!
There is a lot of latency on GitHub CI, especially on macOS and Windows targets. So I had to add some margin on many time-based tests.
I will squash-merge this PR if it seems ok to you. Unfortunately, I had to do many small fixes as the tests failed on the CI since they all worked perfectly locally.
The good news all tests pass on linux/macos/windows with 5 different python interpreters

.github/workflows/pythonpackage.yml
```diff
@@ -43,9 +42,15 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -r requirements.txt
```
dbrattli, line 46: It would be nice to have a requirements.txt that contains the deps needed for development e.g pytest, coveralls etc. and not have them internal to the particular CI.

MainRo: Yes, I added a requirements_dev.txt that contains all test/dev dependencies.

dbrattli: Why not just call it requirements.txt? They should be used for development ref: https://stackoverflow.com/a/43659126 We should also enable `pip install -r requirements.txt` here and remove the individual pip installs in each step.

MainRo: I always used requirements.txt to pin or put constraints on the project dependencies. And other requirement files for dev/doc dependencies. Something like this (a sub-link from your so reference):
https://stackoverflow.com/questions/17803829/how-to-customize-a-requirements-txt-for-multiple-environments/20720019#20720019
But I am not sure what it the good way to do. I just the name to requirements.txt, that's fine for me.

> We should also enable pip install -r requirements.txt here and remove the individual pip installs in each step.

I already removed all "pip  install" and replaced them with one requirements install on each job. Do you mean something else ?

dbrattli: Thanks. It look good now!

dbrattli (approved):
Great work. Well done!

erikkemperman (approved):
This looks good to me, but I should add I am still unfamiliar with GH CI... Shame about the little benchmark I had for Travis, but maybe it wasn't that useful anymore (at the time it demonstrated some performance wins in scheduler refactoring).

MainRo:
> Shame about the little benchmark I had for Travis, but maybe it wasn't that useful anymore (at the time it demonstrated some performance wins in scheduler refactoring).

We should probably see how to add benchmark tests via dedicated platforms/services. This would be great to follow the performances of each release. But considering the indeterministic performances of Github's runners, this has to be done somewhere else.

PR #555, opened 2021-02-02
fix scheduler forwarding in delay operator

timonviola:
Fix a bug where the scheduler was not assigned, when it was provided as an input
to the operator.


MainRo (approved):
Thanks!

PR #564, opened 2021-03-29
Fix using scheduler argument for delay

kormang:
When calling `delay` with `scheduler` argument, it cased items to drop if scheduler has its own `now` property because timestamps where obtained from another scheduler.

### DIFF
--- a/rx/core/operators/delay.py
+++ b/rx/core/operators/delay.py
@@ -92,7 +92,7 @@ def observable_delay_timespan(source: Observable, duetime: typing.RelativeTime,
                     mad.disposable = _scheduler.schedule_relative(duetime, action)
         subscription = source.pipe(
             ops.materialize(),
-            ops.timestamp()
+            ops.timestamp(scheduler=_scheduler)
         ).subscribe_(on_next, scheduler=_scheduler)
 
         return CompositeDisposable(subscription, cancelable)


kormang:
@MainRo @dbrattli PTAL

It's a small bug fix.

MainRo:
can you provide a sample code (or even better an additional unit test) that reproduces the issue? I did not have time to reproduce the issue yet. In the current implementation, the scheduler used in delay should be propagated up to the timestamp operator.

kormang:
OK, I'll need to check it again, because I just saw this commit: https://github.com/ReactiveX/RxPY/commit/fdfe05758028f80663a5a3ddaabe559e37c8f28c

which was not there when I had discovered this bug. Maybe it did fix it. If not, I'll write unit test.

kormang:
O yes, you are right. #555 fixes the problem.

For a reference, here's the code I used to debug this problem:

```python
import rx.operators as ops
from rx.scheduler.eventloop import AsyncIOScheduler
from rx.scheduler.threadpoolscheduler import ThreadPoolScheduler
import asyncio
import logging

from rx import of

logging.basicConfig(
    level=logging.DEBUG,
    format='[%(threadName)s|%(levelname).5s|%(name).32s:%(lineno)d] %(message)s'
)

logger = logging.getLogger('repro')

main_loop = asyncio.get_event_loop()

scheduler = AsyncIOScheduler(main_loop)
tpscheduler = ThreadPoolScheduler(1)

obs = of('FIRST', 'SECOND')

# This part if confusing.
main_loop.call_later(1.0, logger.info, 'Direct schedule 1')
obs = obs.pipe(ops.delay(1.0, scheduler=scheduler))
obs.subscribe(logger.info)
main_loop.call_later(1.0, logger.info, 'Direct schedule 2')

main_loop.run_forever()
```

If you revert #555 , I'll see that it does not print messages FIRST and SECOND.

kormang:
Sorry for bothering you.

kormang:
@MainRo Can you, please, tell me what you think about issue #565 ?

PR #567, opened 2021-04-03
Fix AsyncIOThreadSafeScheduler deadlock

### LINKED ISSUE #566: Deadlock in AsyncIOThreadSafeScheduler
Fallowing code produces deadlock:

```python
import rx
import rx.operators as ops
from rx.subject.subject import Subject
from rx.scheduler.eventloop import AsyncIOThreadSafeScheduler
import asyncio
import logging

def print_threads_traces(_):
    import traceback
    print("\n*** STACKTRACE - START ***\n")
    code = []
    for threadId, stack in sys._current_frames().items():
        code.append("\n# ThreadID: %s" % threadId)
        for filename, lineno, name, line in traceback.extract_stack(stack):
            code.append('File: "%s", line %d, in %s' % (filename,
                                                        lineno, name))
            if line:
                code.append("  %s" % (line.strip()))

    for line in code:
        print(line)
    print("\n*** STACKTRACE - END ***\n")

logging.basicConfig(
    level=logging.DEBUG,
    format='[%(threadName)s|%(levelname).5s|%(name).32s:%(lineno)d] %(message)s'
)

logger = logging.getLogger('repro')

scheduler = AsyncIOThreadSafeScheduler(loop)

src = Subject()

def log_and_push_another_one(msg):
    logger.info(msg)
    src.on_next(msg + "0")
    logger.info('Done')

# Two delays are needed to break the cycle of action self-rescheduling in delay.py.
# That will cause setting cancelable.disposable = mad, which will cause previous
# disposable to be disposed, which will cause scheduler to wait for confirmation,
# which causes deadlock.
obs = src.pipe(ops.delay(1.0, scheduler=scheduler), ops.delay(1.0, scheduler=scheduler))
obs.subscribe(log_and_push_another_one)

# After few seconds print stack trace of each thread, in another thread.
rx.of(1).pipe(ops.delay(6.0)).subscribe(print_threads_traces)

loop.call_soon(src.on_next, "Message0")

loop.run_forever()

```

This is where thread waits:

```
File: "main.py", line 58, in <module>
  loop.run_forever()
File: "/usr/lib/python3.6/asyncio/base_events.py", line 438, in run_forever
  self._run_once()
File: "/usr/lib/python3.6/asyncio/base_events.py", line 1451, in _run_once
  handle._run()
File: "/usr/lib/python3.6/asyncio/events.py", line 145, in _run
  self._callback(*self._args)
File: "RxPY/rx/scheduler/eventloop/asynciothreadsafescheduler.py", line 87, in interval
  sad.disposable = self.invoke_action(action, state=state)
File: "RxPY/rx/scheduler/scheduler.py", line 103, in invoke_action
  ret = action(self, state)
File: "RxPY/rx/core/operators/delay.py", line 69, in action
  result.accept(observer)
File: "RxPY/rx/core/notification.py", line 37, in accept
  return self._accept_observer(on_next)
File: "RxPY/rx/core/notification.py", line 99, in _accept_observer
  return observer.on_next(self.value)
File: "RxPY/rx/core/observer/autodetachobserver.py", line 26, in on_next
  self._on_next(value)
File: "mamicu.py", line 43, in log_and_push_another_one
  src.on_next(msg + "0")
File: "RxPY/rx/subject/subject.py", line 55, in on_next
  super().on_next(value)
File: "RxPY/rx/core/observer/observer.py", line 26, in on_next
  self._on_next_core(value)
File: "RxPY/rx/subject/subject.py", line 62, in _on_next_core
  observer.on_next(value)
File: "RxPY/rx/core/observer/autodetachobserver.py", line 26, in on_next
  self._on_next(value)
File: "RxPY/rx/core/operators/materialize.py", line 23, in on_next
  observer.on_next(OnNext(value))
File: "RxPY/rx/core/observer/autodetachobserver.py", line 26, in on_next
  self._on_next(value)
File: "RxPY/rx/core/operators/map.py", line 41, in on_next
  obv.on_next(result)
File: "RxPY/rx/core/observer/autodetachobserver.py", line 26, in on_next
  self._on_next(value)
File: "RxPY/rx/core/observer/autodetachobserver.py", line 26, in on_next
  self._on_next(value)
File: "RxPY/rx/core/operators/delay.py", line 55, in on_next
  cancelable.disposable = mad
File: "RxPY/rx/disposable/serialdisposable.py", line 39, in set_disposable
  old.dispose()
File: "RxPY/rx/disposable/multipleassignmentdisposable.py", line 47, in dispose
  old.dispose()
File: "RxPY/rx/disposable/compositedisposable.py", line 68, in dispose
  disp.dispose()
File: "RxPY/rx/disposable/disposable.py", line 41, in dispose
  self.action()
File: "RxPY/rx/scheduler/eventloop/asynciothreadsafescheduler.py", line 110, in dispose
  future.result()
File: "/usr/lib/python3.6/concurrent/futures/_base.py", line 427, in result
  self._condition.wait(timeout)
File: "/usr/lib/python3.6/threading.py", line 295, in wait
  waiter.acquire()
```

It does not have to be two delays in a row, it can be observe_on after delay, or anything that prevents next delay action to be rescheduled immediately (see comment in code example).

It may seem to you that it is not real world use case to push item into observable from the same thread and asyncio loop as observer, but it is really important sometimes to be able to do everything in one thread, including pushing new items sometimes.

It may seem to you that we should use basic  AsyncIOScheduler, but if sometimes we still need to do something from other threads, it becomes impossible, I have tried.

To me it seems that scheduler shouldn't really wait for that future, because:
1. Canceling in asyncio is best effort anyway.
2. Other scheduler don't seem to behave like that.
3. If I remove future, all tests pass anyway.

kormang:
Fixes #566 .

If you think about it, canceling is best effort anyway, and there is on guaranties that it will be canceled.

If you're calling dispose from the same thread, you get stuck in deadlock.

If you are calling dispose from another thread, you need to count on the fact, that you might be late, already. So, even with waiting for future result, you have to have in mind that nothing actually got cancelled. It also seems that nobody is even expecting this waiting for future result, since no tests are actually checking it.

It really gets easy to stuck in a dead lock, if you have multiple threads, and one of them runs loop, and it is difficult to avoid both deadlock and _RuntimeError: Non-thread-safe operation invoked on an event loop other than the current one._

Being able to use AsyncIOThreadSafeScheduler from both current thread  and other threads, otherwise it not really thread **safe**.

### DIFF
--- a/rx/scheduler/eventloop/asynciothreadsafescheduler.py
+++ b/rx/scheduler/eventloop/asynciothreadsafescheduler.py
@@ -50,7 +50,7 @@ class AsyncIOThreadSafeScheduler(AsyncIOScheduler):
         handle = self._loop.call_soon_threadsafe(interval)
 
         def dispose() -> None:
-            if not self._is_loop_running_on_another_thread():
+            if self._on_self_loop_or_not_running():
                 handle.cancel()
                 return
 
@@ -107,7 +107,7 @@ class AsyncIOThreadSafeScheduler(AsyncIOScheduler):
                 except Exception:
                     pass
 
-            if not self._is_loop_running_on_another_thread():
+            if self._on_self_loop_or_not_running():
                 do_cancel_handles()
                 return
 
@@ -142,9 +142,14 @@ class AsyncIOThreadSafeScheduler(AsyncIOScheduler):
         duetime = self.to_datetime(duetime)
         return self.schedule_relative(duetime - self.now, action, state=state)
 
-    def _is_loop_running_on_another_thread(self):
+    def _on_self_loop_or_not_running(self):
+        """
+            Returns True if either self._loop is not running, or we're currently
+            executing on self._loop. In both cases, waiting for a future to be
+            resolved on the loop would result in a deadlock.
+        """
         if not self._loop.is_running():
-            return False
+            return True
         current_loop = None
         try:
             # In python 3.7 there asyncio.get_running_loop() is prefered.
@@ -154,4 +159,4 @@ class AsyncIOThreadSafeScheduler(AsyncIOScheduler):
             # thread, we get error like:
             # RuntimeError: There is no current event loop in thread 'Thread-1'
             pass
-        return self._loop != current_loop
+        return self._loop == current_loop
--- a/rx/scheduler/eventloop/asynciothreadsafescheduler.py
+++ b/rx/scheduler/eventloop/asynciothreadsafescheduler.py
@@ -50,10 +50,14 @@ class AsyncIOThreadSafeScheduler(AsyncIOScheduler):
         handle = self._loop.call_soon_threadsafe(interval)
 
         def dispose() -> None:
+            future: Future = Future()
+
             def cancel_handle() -> None:
                 handle.cancel()
+                future.set_result(0)
 
             self._loop.call_soon_threadsafe(cancel_handle)
+            future.result()
 
         return CompositeDisposable(sad, Disposable(dispose))
 
@@ -92,14 +96,18 @@ class AsyncIOThreadSafeScheduler(AsyncIOScheduler):
         handle.append(self._loop.call_soon_threadsafe(stage2))
 
         def dispose() -> None:
+            future: Future = Future()
+
             def cancel_handle() -> None:
                 try:
                     handle.pop().cancel()
                     handle.pop().cancel()
                 except Exception:
                     pass
+                future.set_result(0)
 
             self._loop.call_soon_threadsafe(cancel_handle)
+            future.result()
 
         return CompositeDisposable(sad, Disposable(dispose))
 
--- a/rx/scheduler/eventloop/asynciothreadsafescheduler.py
+++ b/rx/scheduler/eventloop/asynciothreadsafescheduler.py
@@ -50,14 +50,10 @@ class AsyncIOThreadSafeScheduler(AsyncIOScheduler):
         handle = self._loop.call_soon_threadsafe(interval)
 
         def dispose() -> None:
-            future: Future = Future()
-
             def cancel_handle() -> None:
                 handle.cancel()
-                future.set_result(0)
 
             self._loop.call_soon_threadsafe(cancel_handle)
-            future.result()
 
         return CompositeDisposable(sad, Disposable(dispose))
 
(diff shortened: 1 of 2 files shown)


kormang:
@erikkemperman You wrote this code, can you please tell me did you do it on purpose or it just seemed right thing to do?
Is this OK?
Can you, please, review?

Thanks.

erikkemperman:
@kormang I did not actually write this code, although it seems that way on github history because I moved this file in a new directory. You might be right about the deadlock, but looks to me like it was a deliberate choice to make the cancel call block until the future is fulfilled. I'm not entirely sure if anything would break if that is removed. @dbrattli Could you take a look at this PR?

kormang:
OK, wait then. I'll try to detect if we're running on the same loop using https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_running_loop, and in such case  call cancel immediately just like AsyncIOScheduler. Not sure if that is possible, but I'll try later today.

kormang:
I've made it wait for future, when possible. Necessary conditions to be able to wait without deadlock are, `loop_` is running, and it belongs to another thread.

MainRo:
@kormang I just commented on #566 

Can you confirm that your need is to cancel subscriptions that are distributed between several asyncio event loops?
In such a case, I agree we must remove the wait for the future (This was not the intended usage of the ThreadsafeScheduler). However I would expect some strange behaviors due to the fact that the disposal becomes asynchronous: You may have some pending tasks that will execute after the call to "dispose" returns.

More generally, RxPy provides some integration with Asyncio, but I would not try to split observable chains between several event loops: The RxPY APIs are all synchronous, and there are some situations there going back and forth between RxPY functions and asyncio couroutines is not possible.

kormang:
@MainRo I'll continue discussion there. Let here be the place to comment this particular PR.

rx/scheduler/eventloop/asynciothreadsafescheduler.py
```diff
@@ -130,3 +141,17 @@ def schedule_absolute(self,
 
         duetime = self.to_datetime(duetime)
         return self.schedule_relative(duetime - self.now, action, state=state)
+
+    def _is_loop_running_on_another_thread(self):
```
MainRo, line 145: can you change the name and switch the result values?
changing it to something like "self._schedule_to_self_loop" seems easier to read for me rather than "not self._is_loop_running_on_another_thread()"

MainRo (changes requested):
That's fine for me. see my comment to make it easier to read (in my opinion).

rx/scheduler/eventloop/asynciothreadsafescheduler.py
kormang: I've changed it to `_on_self_loop_or_not_running`, because if the loop is not running yet, it would also lead to deadlock.
Hope you don't mind being more clear about what function does for the cost of being longer to read?

I've also added one more test for non-relative schedule and cancel, and deduplicated test code.

MainRo (approved):
Excellent, Thank you!

PR #570, opened 2021-04-21
Set daemon attribute instead of using setDaemon method that was deprecated in Python 3.10

### LINKED ISSUE #569: threading.Thread.setDaemon has been deprecated in favor of setting daemon attribute directly in Python 3.10
Ref : python/cpython#25174

Timer class inherits from threading.Thread

```
rg -t py -w 'currentThread|notifyAll|activeCount|isDaemon|setDaemon'                                      
rx/scheduler/timeoutscheduler.py
51:        timer.setDaemon(True)
86:        timer.setDaemon(True)
```

tirkarthi:
Fixes #569


MainRo (approved):
Thanks!

Issue #93, opened 2016-05-01
How can I subscribe to an Observable on a remote machine using RxPy?


mchen402:
[This can be achieved using IObservable.Remotable using the .NET flavour of Rx](http://stackoverflow.com/questions/3202545/can-reactive-extensions-rx-be-used-across-process-or-machine-boundaries), but what about in Python with RxPy?

dbrattli:
Not yet. The problem is that there are so many ways to do this, so RxPY does not support Remotable. But it should be quite straight forward to use a distributed message queue such as ZMQ or Azure service bus or similar. The problem arises when you want to submit queries remotely using Qbservables. I've done some experimenting in a private branch https://github.com/dbrattli/RxPY/blob/feature/provider/rx/observablequery.py

But actually I'm just waiting for IRP to be open source: http://www.infoq.com/presentations/cloud-rx

mchen402:
Thanks Dag, I took a look at your private branch. However, I'm not so familiar with the `ast` or `bonsai` libraries and feel like I'm missing the point of `ObservableQuery` and `Qbservables`.
 - I see that `Qbservables` are just like `Observables` but all arguments of linq operators that take lambdas now take lambdas as a string. I.e. `observable.map(lambda x: x + 1)` vs `qbservable.map("lambda x: x + 1")` -- what is the point of doing this? Is this to overcome the fact that lambdas cannot be pickled?
 - From what I gather, the idea is to have `qbservable` versions of all linq operators that is sent to and executed on a remote server rather than locally. Is that right?
 - How are messages from a server transmitted to a client? For instance, let's say the server wants to tell a client that an observable has completed. How does it do this?

It'd be very helpful for those like me who know a bit of [`twisted` ](https://twistedmatrix.com/trac/) (but not `ast` or `bonsai`)  if you could knock up a minimal working example of how to use `Qbervables` with `twisted`. One example might be an Rx echo server: the client sends `'Hello World'` to the server and the server response with something like `observable.from_iterable('Hello World')` which the client can subscribe to.

> But actually I'm just waiting for IRP to be open source: http://www.infoq.com/presentations/cloud-rx

[...]

Issue #70, opened 2016-01-02
ProcessPoolScheduler


ChromoX:
Hello,

I am currently using Reactive Programming paradigm as an alternative to the Actor model. I was expecting that it would be easy to dictate processing in multiple processes by simply using the correct scheduler. After looking I can't find any scheduler that seems to use multiple processes. I was thinking about building a ProcessPoolScheduler, but I was wondering what problems you guys see with implementing something like that?

I understand that the Reactive contract is such that streams are processed in order, but I know many applications don't require that assumption.

Also if something like the ProcessPoolScheduler is not possible, how might one go about trying to build a parallel processing system in Python where you have the GIL to contend with.

Thanks,
Jeffrey

dbrattli:
This is a very interesting topic. There have been some previous discussions in #65 and #67 that you should look at. Some initial thoughts: There is nothing in the Rx contract that prevents you from processing items in parallel, as long as ordering is maintained (think overlapping processing). If ordering is not relevant for your application then you should use the initial sequence to just schedule work, and have that create new streams of processed work that you can flatten to a result stream (see #67 for an example).

You also need to check if the work you want to do in parallel is IO or CPU bound and if the processing is done by extensions libraries or by Python code. If the work you want to do is IO bound, you should probably use `AsyncIOScheduler` to achieve concurrency. If work is CPU bound but done in extension libraries like NumPy, then use the `ThreadPoolScheduler`. Work will be done in parallel since extension libraries like NumPY are not affected by the GIL. If work is CPU bound, and work needs to be done by Python code, then the GIL is a problem and you may want to write a `ProcessPoolScheduler`. If a `ProcessPoolScheduler` is the right answer or not depends on how long-running your CPU bound work will be.

But I can't see any problems with building a `ProcessPoolScheduler` so it would be awesome if you want to try. Then the next step could be to make an `RemotingScheduler` to schedule work on other nodes.

ChromoX:
Thanks for the response. I've been trying to wrap my head around how invoke/disposables work in schedulers. Regardless I assume that the RxPy project doesn't have a ThreadPoolScheduler? I couldn't find one in the rx.concurrency folder.

dbrattli:
No we don't have a `ThreadPoolScheduler` yet either, so there are many possible projects for those who want to contribute.

tadas-subonis:
Did anybody do any work on this?

MainRo:
For information I published a [ray](https://ray.io) operator here:
[...]

Issue #45, opened 2015-04-19
(idea) Is it crazy to integrate RxPy and Celery, or RxPy and RabbitMQ?


hangtwenty:
Just an idea, that I'm wondering if I should explore or nip in the bud. Looking for some Rx wisdom I suppose, since I'm a newb.

I'm looking at these:
- http://pika.readthedocs.org/en/0.9.14/examples/twisted_example.html (Since there's an asyncio scheduler there can probably be a Twisted scheduler ... [right](https://glyph.twistedmatrix.com/2014/05/the-report-of-our-death.html)?)
- http://pika.readthedocs.org/en/0.9.14/examples/tornado_consumer.html

I'm also thinking about [Celery](http://celery.readthedocs.org/en/latest/userguide/canvas.html) and how to either integrate it with RxPy, or to imitate a small part of RxPy with Celery's operators. Integration would be nicer. Or if RxPy + ((Twisted or Tornado) + pika + RabbitMQ) could replace the need for Celery for some projects... that would also be interesting.

Is this something worth thinking about and trying to work on? Or am I mixing incompatible patterns? Opinions?

dbrattli:
Are you trying to build something similar to [Storm](http://storm.apache.org) like a real-time processing flow/network, or more like the [Akka](http://akka.io/) actors for highly concurrent distributed applications?

hangtwenty:
Great question.

I found my way to ReactiveX via Akka and [reactive-streams](http://www.reactive-streams.org/). (Specifically, [this presentation](http://spray.io/msug/#/).) So Akka is relevant, but I'm not necessarily after that actor model. My attention was caught by Akka's DSL for reactive streams, a DSL that looks a lot like RxJava.

I'm after the "highly concurrent" and "distributed" aspects, and the rest is unclear. My primary imagined use-case is very IO-bound (within workers). Perhaps RxPY the wrong tool for this kind of job.

Regarding Storm, the [streamparse](https://github.com/Parsely/streamparse) library is compelling. And it's README says this (*emphasis* mine):

> It can be viewed as a more robust *alternative to Python worker-and-queue systems, as might be built atop frameworks like Celery and [RabbitMQ].* It offers a way to do "real-time map/reduce style computation" against live streams of data. It can also be a powerful way to scale long-running, highly parallel Python processes in production.

[...]

PR #572, opened 2021-05-08
[WIP] Add multi-processing and distributed computing support via ray

MainRo:
This is a proposal to add multi-processing and distributed computing support to RxPy.
The feature is exposed via a *distribute* operator and uses [ray](https://ray.io/) as a backend.
I chose to go with an operator rather than a scheduler for several reasons:

- The scheduler API does not match the API of frameworks like ray.
- Using an operator taking the remote pipeline as a parameter seems easier to use.

The *distribute* operator takes a pipeline as an input parameter. This pipeline is executed on remotes processes or remote nodes, with the source item of the operator being distributed to the workers. The operator has other arguments to configure the number of workers and how to distribute the items. All items emitted by the workers are sent back to the original operator. The default configuration preserves the order of the processed items as if they were serialized.

The usage looks like this:

```
source.pipe(
    rxray.distribute(rx.pipe(
        ... some heavy computation pipe, parallelized on worker nodes ....
    )),
   ... the output items of all workers are centralized here for further processing if needed...
).subscribe()
```

This is a proposal that could solve the following issues: #70 #93 #45

Pros of adding this operator:

- A simple way to address both multi-processing and distributed computing.
- Ray can be used as a local library (multiprocessing is used), or as a cluster (distributed computing)
- Ray is a very popular python solution for parallelization.

Cons of adding this operator

- Adds a dependency on Ray: Other people may/will request support to other similar frameworks (dask, celery...). How would we deal with such requests?
- The current implementation is very small, but I may evolve in a complex way depending on the users' needs: Distributed computed is complex and covers many different needs.

Limitations:
- No support of higher-order observables

There is still some cleanup to do on this code and add documentation. I tested it on a single machine, and I am working on tests on a Kubernetes cluster.

What is your opinion on this? Should we add it to RxPy or should it be part of a third-party package?
@dbrattli @erikkemperman @jcafhe

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -61,6 +61,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -r requirements.txt
+          pip install ray
       - name: Run coverage tests
         run: |
           coverage run -m pytest
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -28,7 +28,28 @@ jobs:
     - name: Test with pytest
       run: |
         pytest
-  
+
+  test_ray:
+    strategy:
+      matrix:
+        platform: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: [3.6, 3.7, 3.8, 3.9, pypy3]
+    runs-on: ${{ matrix.platform }}
+
+    steps:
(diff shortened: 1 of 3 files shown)


MainRo:
Closing this PR, I published a dedicated package here:
https://github.com/maki-nage/rxray

I think it makes more sense because it has a *big* dependency (ray).

PR #575, opened 2021-06-27
Remove deprecated loop parameter

carlwgeorge:
asyncio.sleep's loop parameter has been deprecated since Python 3.8, and is [removed in 3.10](https://docs.python.org/3.10/library/asyncio-task.html#sleeping).  The tests fail on 3.10 unless this is removed.


MainRo (approved):
Thanks!

PR #580, opened 2021-08-29
Correct example code instructions

Julian-O:
Python's `input()` function does not return when *any* key is pressed. The `Enter` key must be pressed (on typical keyboards).


MainRo (approved):

MainRo:
Thanks!

Issue #578, opened 2021-08-04
Zip operator not working


OopsYao:
The demo of `zip` operator in the [migration guide](https://rxpy.readthedocs.io/en/latest/migration.html#removal-of-the-result-mapper) does not work as expected.
```python
import rx
from rx import operators as ops
import operator

a = rx.of(1, 2, 3, 4)
b = rx.of(2, 2, 4, 4)

a.pipe(
    ops.zip(b), # returns a tuple with the items of a and b
    ops.map(lambda z: operator.mul(z[0], z[1]))
).subscribe(print)
```
Expected: print 2, 4, 12, 16 sequentially; actually there is no output.

OS: Arch Linux x86_64 with kernel 5.13.6-arch1-1
Python: 3.9.6
Rx: 3.2.0

noflame:
The same thing happened to me here.
So I had to use version 3.1.1, the zip function works fine

os: Win10
Python 3.7.6

elbowz:
+1

MainRo:
Fixed in #595

chunzhao:
Noob question: Is the fix of this issue going to be patched into RxPY v3.0? 
I'm relying on pip to install RxPY (pip3 install rx). The issue of zip operator exists in v3.2.0 as of today.

HungryAfterLunch:
Some here. The zip operator work fine in v3.1.1 but not working on v3.2.0.

os: win11
python: 3.10.4

Issue #525, opened 2020-06-26
zip operator should complete if a single upstream source completes


MichaelSchneeberger:
The following observable `y` ...

``` python
import rx
from rx import operators as rxop

y = rx.range(2).pipe(
    rxop.zip(rx.defer(lambda _: y)),
    rxop.merge(rx.just(0)),
    rxop.share(),
)

y.subscribe(print, on_completed=lambda: print('completed'))
```

... never completes.

My suggestion: The zip operator should complete if any upstream source completed and the corresponding queue is empty.

That is how it is implemented in [RxJava](https://github.com/ReactiveX/RxJava/blob/3.x/src/main/java/io/reactivex/rxjava3/internal/operators/observable/ObservableZip.java).

MainRo:
I agree that zip should complete in such cases, because no item will be emitted after one source completes. Some applications may use the current behavior to wait for all streams to complete, but probably most people expect it to complete immediately.

hoc081098:
Hi @MichaelSchneeberger, I created a PR #537

PR #595, opened 2022-01-22
rework zip fix for #537

MainRo:
The original fix for #525 breaks sequences where some observables emit
item faster than other ones. By completing too soon, the remaining
observables cannot catchup later. A very simple case is in #578 where
the two observables to zip emit their items sequentially.

We can fix both issues by completing whenever an observable completes
and there is no queued item. Otherwise we let the remaining observables
a chance to emit new items before completion.

### DIFF
--- a/rx/core/observable/zip.py
+++ b/rx/core/observable/zip.py
@@ -34,6 +34,7 @@ def _zip(*args: Observable) -> Observable:
         n = len(sources)
         queues: List[List] = [[] for _ in range(n)]
         lock = RLock()
+        is_completed = [False] * n
 
         @synchronized(lock)
         def next(i):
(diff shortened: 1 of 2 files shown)


MainRo:
@MichaelSchneeberger can you review this?

dbrattli (approved):
Looks good!

MichaelSchneeberger:
I would have done it like this
```
def next(i):
    if all([len(q) for q in queues]):
        try:
            queued_values = [x.pop(0) for x in queues]
            res = tuple(queued_values)
        except Exception as ex:  # pylint: disable=broad-except
            observer.on_error(ex)
            return
  
        observer.on_next(res)
  
        # after sending the zipped values, complete the observer if at least one upstream observable
        # is completed and its queue has length zero 
        if any((done for queue, done in zip(queues, is_completed) if len(queue)==0)):
            observer.on_completed()

def completed(i):
    is_completed[i] = True

    if len(queues[i]) == 0:
        observer.on_completed()
```

MainRo:
Thank @MichaelSchneeberger, your proposal is simpler and more readable. I updated the PR with it.

PR #596, opened 2022-01-26
Removal of misleading information in take_while documentation

kkonrad:
Small thing. Documentation was stating that `element's index is used in logic of the predicate function` which is not true for `take_while` operator


MainRo (approved):
Thanks!

PR #597, opened 2022-02-02
Generic type annotations

### LINKED ISSUE #461: Add generic typings
with python3.5+, we now have [static-typing](https://docs.python.org/3/library/typing.html) with [generics](https://docs.python.org/3/library/typing.html#generics)

But the current code inside rxpy does not account for generics:

```python
class BehaviorSubject(Subject):
    """Represents a value that changes over time. Observers can
    subscribe to the subject to receive the last (or initial) value and
    all subsequent notifications.
    """

    def __init__(self, value) -> None:
        ...
```

One quick fix for this is adding `typevars`:

```python
from typing import (
    Generic,
    TypeVar,
    Callable,
)

T = TypeVar('T')


class Subject(Generic[T]):
    def subscribe(self, func: Callable[[T], None]):
        ...


U = TypeVar('U')


class BehaviorSubject(Subject[U], Generic[U]):
    def __init__(self, value: U):
        ...
```

Note that current IDE support is not "very good" like typescript.
But that'll improve, as everyone (google/fb/ms/dropbox/etc) is trying to create a decent type-checker

### LINKED ISSUE #514: Type hint support question
I am running 
```
Python 3.8.2
```

```pip show rx
Name: Rx
Version: 3.1.0
Summary: Reactive Extensions (Rx) for Python
Home-page: http://reactivex.io
Author: Dag Brattli
Author-email: dag@brattli.net
License: MIT License
Location: XXX
Requires:
Required-by:
```

I would like to use the rx library using python type hints. My very simple attempt:

```py
from unittest.case import TestCase
from rx.subject import BehaviorSubject
from rx import Observable


class TestCounter(TestCase):

    def test_typing(self):

        sub = BehaviorSubject[int](1)

        sub.on_next(10)

        ob: Observable[int] = sub
```

This does not run and fails with:
```
TypeError: <class 'rx.subject.behaviorsubject.BehaviorSubject'> is not a generic class
```

When I remove the generics (`[int]`) from the type hints then the code runs but my type checker does not detect that `BehaviorSubject` is actually a subject of integers.  Also it would not detect that `ob` is an observable of integers.

The other difficulty I have is when trying to declare a type that represents an operator function. To my best knowledge this should be:

```py
from typing import Callable, TypeVar

from rx import Observable

T1 = TypeVar('T1')
T2 = TypeVar('T2')
Operator = Callable[[Observable[T1]], Observable[T2]]
```

I would expect this to declare a function that accepts an observable of type `T1` and returns an observable of type `T2`.
But running this gives:

```
TypeError: <class 'rx.core.observable.observable.Observable'> is not a generic class
```

I also tried to import `Observable` from `rx.core.typing` since that version is generic. But this does not help either, because that version of `Observable` does not define the `pipe` method. Also a deep import does not seem to be appropriate.

Could you please point me to examples that show how to use RxPY with type hints and proper generics support.

### LINKED ISSUE #531: Mypy reports "Module 'rx' has no attribute 'Observable'" since 3.1.1
I don't know what has changed, but after I updated to RxPy 3.1.1, mypy began to report such errors as "Module 'rx' has no attribute 'Observable'", possibly due to python/mypy/issues/7423.

It happens for all other major types such as `BehaviorSubject`, and changing `[mypy] implicit_reexport` in `setup.cfg` to `True` fixes the symptom.

### LINKED ISSUE #540: Generic rx.core.typing.Observable[T] does not include .pipe()
For example, using `.pipe`:

```python
from rx import Observable
from rx.core.typing import Observable as ObservableT

def func(observable: ObservableT[int]):
    return observable.pipe()
```

`mypy` complains with:

```
test.py:5: error: "Observable[int]" has no attribute "pipe"
Found 1 error in 1 file (checked 1 source file)
```

### LINKED ISSUE #586: Clarity on the correct import for `Observable`
Documentation for `Observable` documents it as `rx.Observable` (https://rxpy.readthedocs.io/en/latest/reference_observable.html#rx.Observable) so I would assume to import it as

```python
from rx import Observable
```

however [pylance ](https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-pylance) reports this as an error:
image

I assume this is because `rx` does not list it as its exports in `__all__` from its `__init__.py`.

What is the intended way to import `Observable`? Would it be possible to list the official exports in `__init__.py` files?

dbrattli:
- Will be part of version 4. We need a major version bump because of Python dependency changes.
- Deprecate Python 3.6 since it ended its life 23. dec 2021
- Add Python 3.10 and add  to CI
- Use modern and generic type-hints (PEP-484) so we can use pylance/pyright at strict settings
- Rename abstract base classes to post-fix with `*Base` to make it easier to distinguish them e.g `ObserverBase` instead of `Observer` which was easy to confuse with `rx.Observer`. At least this made this refactor a lot easier.
- Refactor subscribe logic to have same signature for base class `ObservableBase` as `Observable`. Removes the `subscribe_()` method.
- Convention: All `abc` types should also be prefixed with `abc.` so we now when we are dealing with them. I.e use `abc.ObserverBase` instead of just `ObserverBase`. 
- Convention: All `typing` types should be prefixed with `typing.` so it's easy to distinguish them. E.g `typing.Mapper`
- Use black formatter and formatting at default settings (it's used by everyone else) so it's easier to read and write the code. Ideally a formatter should be a separate PR, but it was a lot of help here.
- Use isort to make sure imports are sorted correctly
- Remove support for Tornado.Future (Tornado now uses asyncio.Future)
- Move statistics operators to examples folder
- TODO: deprecate `from_callback`. It's a JavaScript artifact ported to Python not part of the original Rx code.

This is the start of a massive rewrite of type hints to get proper generic type annotations. The code itself has not been changed other than merging `types.py` with `abc` types so that `abc` is now typed. No reason anymore for having untyped base classes. This will enable us to use [pylance/pyright](https://github.com/microsoft/pyright) for static type checking in addition to mypy. The goal is to eventually add mypy/pyright to CI if we can eventually get all of the code to type check.

It's probably better to merge this before all operators have been rewritten, i.e do this is multiple stages as other PR's will mess up this with conflicts that may make the job impossible to complete.

- Fixes #461
- Fixes #514 
- Fixes #531
- Fixes #540 
- Fixes #586


erikkemperman:
@dbrattli Hi Dag -- this sounds like an excellent plan! Unfortunately I am pretty busy at the moment, so I don't feel I could do this work justice in terms of review... I hope you understand!

dbrattli:
@erikkemperman Yes, understand. I've know all about being busy with other stuff. This PR is btw pretty much impossible to review (in detail) so what I'm looking for is an "approved in principle" that this is the way forward. The API should be exactly the same so it's just the added types to avoid static type checkers to choke when analyzing the code. FYI: @MainRo

MainRo:
I will look at it. I will try to (partially :) read it in the coming days.

dbrattli:
FYI: Longer pipelines like this now actually type checks at with Pylance at strict settings 🎉 😊 

Screenshot 2022-02-09 at 08 48 09

MainRo:
If we go for a major version bump I would also consider other evolutions:

Removing some of the schedulers: Most of them are not really maintained I think. So I would remove them, and eventually let interested people move them to dedicated repos (or agree to be maintainers on these parts). I had to work on a few bug reports for some of them and it was not fun at all (learning a framework that I will probably never use...). tk, pygrame, wx, and eventlet would be on my removal list.

Adding kwargs on subscribe: Since v3, the scheduler is backpropagated during subscription. Several times I would have needed the possibility to send other arguments at subscription time to sources or custom operators. We could either add a kwargs argument to subscribe, and also make subscribe part of these kwargs.

I started to look at the changes. overall, it looks good.

We can create a dedicated issue to track all v4 changes, like we did with v3.

dbrattli:
Yes, that sounds good @MainRo . Let's create a v4 issue to discuss other enhancements and deprecations. Then do pre-releases for anyone interested until we are happy with the result and can do a proper release.

dbrattli:
@MainRo This PR is ready to be merged. Can I get a signoff here? There's more to be done, but I don't want to do it here. This PR now type-checks fully with pyright. Mypy cannot keep up with the pipe-operator, but public API checks for mypy as well.

MainRo:
@dbrattli you closed this PR without merging ?

dbrattli:
Yes, I got blocked waiting for sign-offs. The PR was impossible to review anyways. So I changed my mind. We don't really need to merge it. It now lives as the `reactivex` branch and I plan to start publishing alphas from it under the `reactivex` package on PyPI. Then new PRs will be against the `reactivex` branch instead of `master` and we don't have to worry about the current master branch.

PR #602, opened 2022-02-20
Fix wxscheduler issues

christiansandberg:
I just saw that you are planning on deprecating these schedulers but since I already made the fixes I'm making a PR anyway.

The Timer class used for scheduling actions has a few drawbacks:

* Timers can't be started from other threads, making operators like `observe_on` useless.
* Timers are quite slow. When doing an immediate schedule it is much faster to use [CallAfter](https://docs.wxpython.org/wx.functions.html#wx.CallAfter) instead.

### DIFF
--- a/reactivex/scheduler/mainloop/wxscheduler.py
+++ b/reactivex/scheduler/mainloop/wxscheduler.py
@@ -77,9 +77,9 @@ class WxScheduler(PeriodicScheduler):
         timer = self._timer_class(interval)
         # A timer can only be used from the main thread
         if self._wx.IsMainThread():
-            timer.Start(msecs, oneShot=not periodic)
+            timer.Start(msecs, oneShot=not periodic)  # type: ignore
         else:
-            self._wx.CallAfter(timer.Start, msecs, oneShot=not periodic)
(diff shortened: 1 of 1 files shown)


dbrattli:
Hi, thanks for the PR. We are discussing if we should deprecate them so things are not decided yet. Activity is always a good sign of not deprecating.

dbrattli (approved):
Thanks for fixing. PS: You might want to also make a PR against the `reactivex` branch that will be the next version of RxPY (code name ReactiveX)

dbrattli:
@christiansandberg Could I ask you to please update your PR against the latest v4 code, and I will make sure the scheduler will be part of the release and not be deprecated.

christiansandberg:
Oh okay, I thought you would merge both to master eventually. I'll make a new PR.

dbrattli:
PS: See instructions here for setting up the development environment: https://github.com/ReactiveX/RxPY/blob/becb44c29e563c261d8e1a34dd94e68144fea481/README.rst#development

christiansandberg:
Ok, I've rebased on top of master now.

dbrattli (approved):

PR #603, opened 2022-02-22
[ReactiveX] Use pyproject.toml and poetry

dbrattli:


pyproject.toml
```diff
@@ -0,0 +1,81 @@
...
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Topic :: Software Development :: Libraries :: Python Modules",
+]
+packages = [
+    { include = "rx" },
+    { include = "rx/py.typed" }
+]
+
+[tool.poetry.dependencies]
+python = ">= 3.7, < 3.11"
```
MainRo, line 31: Does this mean a release with this will not install on python > 3.10?
If yes, this may be too conservative.

rx/_version.py
```diff
@@ -0,0 +1,2 @@
+# Please make sure the version here remains the same as in project.cfg
```
MainRo, line 1: it is now project.toml instead of project.cfg

pyproject.toml
dbrattli: It says that it needs that be less than 3.11. That includes 3.10, but does not include 3.11 (which is in alpha-5)

dbrattli: We can enable it if someone tries 3.11 and confirms it works with

PR #634, opened 2022-03-13
Add switch_map operator and equivalent starred and indexed

hamstap85:
This draws from the definition in rxjs and maintains parity with the
map operator and its variants


reactivex/operators/__init__.py
```diff
@@ -3325,6 +3326,202 @@ def switch_latest() -> Callable[
...
+
+    Returns:
+        A partially applied operator function that takes an observable
+        source and returns an observable sequence whose elements are
+        each element of the result of invoking the transform function
+        on each element of the source.
+    """
+    from ._switch_map import switch_map_
+
+    return switch_map_(mapper)
```
dbrattli, line 3358:
```suggestion
    from ._switchlatest import switch_latest_

    mapper_: typing.Mapper[_T1, Union[Future[_T2], Observable[_T2]]] = mapper or cast(
        Callable[[_T1], Observable[_T2]], of
    )
    return compose(
        map(mapper_),
        switch_latest_(),
    )
```

dbrattli: So I don't think we need the full (re)implementation of `switch_map` and we should just reuse `map` and `switch_latest`. This is why is this operator is questionable to add since it's so easy to compose yourself. But I think adding it will help to show that.

hamstap85: Interesting, I hadn't really noticed what `switch_latest` was doing. There's so many operators to keep track of.

`mypy` wasn't happy until I made the type value of the cast the same as the annotation of `mapper_`, which makes sense.

dbrattli:
The main building blocks of Rx is the merge functions (`merge`, `switch_latest`, `concat`) combined with the mapper forms the functions we know as `flat_map` (i.e [map_merge](https://github.com/ReactiveX/RxPY/blob/master/reactivex/operators/_flatmap.py#L34)), `switch_map` and `concat_map`.

tests/test_observable/test_switch_starmap.py
```diff
@@ -0,0 +1,437 @@
...
+
+
+# Helper function for raising exceptions within lambdas
+
+
+def _raise(ex):
+    raise RxException(ex)
+
+
+class TestSelect(unittest.TestCase):
```
dbrattli, line 29: Rename

tests/test_observable/test_switch_map.py
```diff
@@ -0,0 +1,500 @@
...
+
+
+# Helper function for raising exceptions within lambdas
+
+
+def _raise(ex):
+    raise RxException(ex)
+
+
+class TestSelect(unittest.TestCase):
```
dbrattli, line 30: Rename

reactivex/operators/__init__.py
```diff
@@ -3325,6 +3326,212 @@ def switch_latest() -> Callable[
     return switch_latest_()
 
 
+def switch_map(
+    mapper: Optional[Mapper[_T1, Observable[_T2]]] = None
+) -> Callable[[Observable[_T1]], Observable[_T2]]:
+    """
+    The switch_map operator.
+
+    Project each element of an observable sequence into a new observable.
```
dbrattli, line 3335:
```suggestion
    Project each element of an observable sequence into a new observable. producing values only from the most 
    recent observable sequence.
```

reactivex/operators/__init__.py
```diff
@@ -3325,6 +3326,212 @@ def switch_latest() -> Callable[
...
+
+
+def switch_map_indexed(
+    mapper_indexed: Optional[MapperIndexed[_T1, Observable[_T2]]] = None
+) -> Callable[[Observable[_T1]], Observable[_T2]]:
+    """
+    The switch_map_indexed operator.
+
+    Project each element of an observable sequence into a new observable
+    by incorporating the element's index.
```
dbrattli, line 3373:
```suggestion
    by incorporating the element's index and producing values only from the most recent
    observable sequence.
```

reactivex/operators/__init__.py
```diff
@@ -3325,6 +3326,212 @@ def switch_latest() -> Callable[
...
+
+def switch_starmap(
+    mapper: Optional[Callable[..., Observable[Any]]] = None
+) -> Callable[[Observable[Any]], Observable[Any]]:
+    """The switch_starmap operator.
+
+    Unpack arguments grouped as tuple elements of an observable sequence
+    and return an observable sequence whose values are each element of
+    the observable returned by invoking the mapper function with star
+    applied on unpacked elements as positional arguments.
```
dbrattli, line 3437:
```suggestion
    applied on unpacked elements as positional arguments, and  only producing values only from the most recent
    observable sequence.
```

reactivex/operators/__init__.py
```diff
@@ -3325,6 +3326,212 @@ def switch_latest() -> Callable[
...
+
+    Args:
+        mapper: A transform function to invoke with unpacked elements
+            as arguments.
+
+    Returns:
+        An operator function that takes an observable source and returns
+        an observable sequence whose values are each element of the
+        observable returned by invoking the mapper function with the
+        unpacked elements of the source.
```
dbrattli, line 3523: Same here: " ... producing values only from the most recent observable sequence.""" or else it looks like a flat_map.

dbrattli:
@hamstap85 One thing that worries me about this PR is that the tests are derived from `test_map` and not e.g `test_switchlatest`. Thus the tests do not have the notion of time and behave more like a regular `flat_map` / `concat_map` in the sense that the switching property of  `switch_map` is not tested. I.e the automatic disposal of the previous observable being subscribed. This is also reflected in the docstrings of the operators where the " producing values only from the most recent observable sequence" is missing. The current docstrings sounds more like a `flat_map` so I'm worried that your original impl also behaved more like a `flat_map` / `concat_map`. Sorry for being difficult, but I just need to be sure that you are implementing what you think you are implementing.

hamstap85:
@dbrattli No worries at all, thanks for the feedback.

What's the desired max line length in docstrings? Equal to python code max length enforced by `black`?

dbrattli:
Docstrings should follow PEP-8, i.e 72 chars per line. PS: all the current tests for `switch_map` and `switch_map_index` pass if you replace `switch_map` with `flat_map` etc. The only difference is that `switch_map` uses `of` instead of `identity` as the default mapper function. But in my new they should either both use `of` or both use `identity`. Do you have a reference for another library using `of` for the default mapper? I'm not opposed for changing the default for `flat_map` but I cannot see how this is currently handled in e.g RxJS. FYI: https://www.baeldung.com/rxjava-flatmap-switchmap

lucasmonstro:
Hello, how can I have the same effect with rxpy?

```typescript
import { map, Subject, switchMap, take, timer } from 'rxjs';

type Point = [number, number];

const DELAY = 1000;
const pixelSequence$ = new Subject<Point[]>();

const mousePoint$ = pixelSequence$.pipe(
  switchMap((sequence) =>
    timer(0, DELAY).pipe(
      map((i) => sequence[i]),
      take(sequence.length)
    )
  )
);

mousePoint$.subscribe(([x, y]) => setMousePosition(x, y));

function setMousePosition(x: number, y: number) {
  console.log(`Mouse set to: [${x}, ${y}]`);
}

const FAKE_SEQUENCE_1: Point[] = [
  [0, 0],
  [10, 10],
  [20, 20],
];

const FAKE_SEQUENCE_2: Point[] = [
  [100, 50],
  [200, 50],
  [300, 50],
];

setTimeout(() => pixelSequence$.next(FAKE_SEQUENCE_1), 0);
setInterval(() => pixelSequence$.next(FAKE_SEQUENCE_2), 1500);
```

PR #639, opened 2022-04-06
Fix missed renaming causing broken docs page

bostikforever:


dbrattli (approved):
Thanks for fixing. We can merge this once the build error has been fixed by https://github.com/ReactiveX/RxPY/pull/640

PR #642, opened 2022-05-08
Fix scheduler not propagated in take_until

christiansandberg:


dbrattli (approved):

PR #644, opened 2022-05-10
fix: make package name all lowercase

### LINKED ISSUE #643: ReactiveX wheel distribution name does not match .dist-info folder name
Using the latest `reactivex` wheel (4.0.0) as an example, the wheel name is:

- `ReactiveX-4.0.0-py3-none-any.whl`

But the dist-info file is named:

- `reactivex-4.0.0.dist-info`

Contrast this with the older `Rx` wheels: 

- `Rx-3.2.0-py3-none-any.whl`
- `Rx-3.2.0.dist-info`

Now this is an extremely minute issue - but it does break the [dependency resolver for Bazel](https://github.com/bazelbuild/rules_python/blob/beed9a257cc493ebf5aa7b8e6bbe577fa637e8f6/packaging/whl.py#L68-L72), which expects the syntax `{distribution}-{version}.dist-info`. 

[PEP 491](https://peps.python.org/pep-0491/#file-contents) somewhat outlines the naming convention for wheel contents relative to its wheel file.

**Additional context**
 - RxPY version 4.0.0

dbrattli:
Fixes #643

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [tool.poetry]
-name = "ReactiveX"
+name = "reactivex"
 version = "0.0.0" # NOTE: will be updated by publish script
 description = "ReactiveX (Rx) for Python"
 readme = "README.rst"


chowder:
Hi @dbrattli , @MainRo , any chance we could get this merged soon? Thanks once again for raising this.

Issue #646, opened 2022-05-19
Use deque in ReplaySubject


timothy-shields:
`ReplaySubject` currently uses a list for its internal queue. Frequently removing items from the front of a list, as is done in the `_trim` method, can be expensive for large lists.

The `ReplaySubject` could be changed to use a `deque` for its queue. This would give faster performance for large replay queues.

It looks like the change would involve:

- initializing `self.queue` to an empty `deque` in `__init__`
- replacing calls to `self.queue.pop(0)` in `_trim` with `self.queue.popleft()`

PR #648, opened 2022-06-05
Use deque in ReplaySubject (#646)

timothy-shields:
See #646 for details.


dbrattli (approved):

PR #654, opened 2022-06-29
Make `_T` in `Observable[_T]` covariant

### LINKED ISSUE #653: Type-var `_T` in `Observable[_T]` should be covariant
The only type-var `_T` of `Obserable` is currently declared as invariant and is causing false positive with the type-checker I'm using (namely Pylance/Pyright).

As we can see, this type-var only describes output values, and it works just like `_T_out` of `rx.abc.ObservableBase`, so there is little reason to make it invariant.

The current typing is clearly forbiding `ops.switch_latest()` being applied to a proper observable. `switch_latest` accepts a `Observable[Observable[_T] | Future[_T]]` as its argument. With `_T` in `Observable[_T]` being invariant, the original observable is required to emit values with type of exactly `Observable[?] | Future[?]`, which is very unlikely.

**To Reproduce**
Let Pylance/Pyright check the following sample code.

**Expected behavior**
The sample code should pass type checks.

**Code or Screenshots**
```python
import reactivex as rx, reactivex.operators as ops

a: rx.Observable[rx.Observable[int]] = globals()[""]
b = a.pipe(ops.switch_latest())
```
image

**Additional context**

 - OS [e.g. Windows]
Doesn't matter.

 - RxPY version [e.g 4.0.0]
 4.0.2

 - Python version [e.g. 3.10.2]
Doesn't matter.

Azureblade3808:
Fixes #653

### DIFF
--- a/reactivex/observable/observable.py
+++ b/reactivex/observable/observable.py
@@ -50,7 +50,9 @@ class Observable(abc.ObservableBase[_T_out]):
 
     def subscribe(
         self,
-        on_next: Optional[Union[abc.ObserverBase[_T_out], abc.OnNext[_T_out], None]] = None,
+        on_next: Optional[
+            Union[abc.ObserverBase[_T_out], abc.OnNext[_T_out], None]
+        ] = None,
         on_error: Optional[abc.OnError] = None,
         on_completed: Optional[abc.OnCompleted] = None,
         *,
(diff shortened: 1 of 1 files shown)


dbrattli (approved):
This look great! Thanks for fixing!

PR #656, opened 2022-07-12
Allow any python in the 3.x series >=3.7

mmv:
Before converting from setup.py to pyproject, this upper bound didn't exist.

There's no reason to cap at 3.10. This breaks installations on other poetry projects defined with Python=^3.7 and would force updating this file for every new python version.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ packages = [
 ]
 
 [tool.poetry.dependencies]
-python = ">= 3.7, < 3.11"
+python = ">= 3.7, < 4.0"
 typing-extensions = "^4.1.1"
 
 [tool.poetry.dev-dependencies]


dbrattli:
Instead of this I suggest that you change `python-package.yaml` to test specifically for a higher python version e.g `3.11.0-beta.4` so we know if it actually works or not. See https://github.com/actions/python-versions. Then we can change pyproject.toml to also include 3.11.

dbrattli:
This makes me unsure if we should accept every new minor version: https://stackoverflow.com/questions/60421663/is-python-3-semantically-versioned-and-forwards-compatible. But I understand the frustration of not being able to use it if it actually works. Parts of me want to approve this PR 😄

dbrattli (approved):
Thanks for fixing

MainRo:
Thanks @dbrattli ! I also felt this restriction was... too restrictive.

PR #666, opened 2022-11-06
Update pyright to 1.1.286

dbrattli:


reactivex/operators/_distinct.py
```diff
@@ -32,7 +32,9 @@ def distinct_(
     key_mapper: Optional[typing.Mapper[_T, _TKey]] = None,
     comparer: Optional[typing.Comparer[_TKey]] = None,
 ) -> Callable[[Observable[_T]], Observable[_T]]:
-    comparer_ = comparer or default_comparer
+    comparer_: typing.Comparer[_TKey] = comparer or cast(
```
MainRo, line 35: Is it really necessary to add a type hint here? Type checkers should infer it since both _compare_ and _default_comparer_ are already defined.

dbrattli: @MainRo Thanks, yes you are right. I updated to pyright v286 and I can now remove the type-hint.

MainRo (approved):

PR #670, opened 2022-12-06
[DOCS] Getting started: fix example

matiboy:
Fix example for custom operator which still uses v3's rx.pipe


dbrattli:
Thanks for fixing. Combining two operators is a composition, not a pipe (pipe applying a source to an operator) so this rename was needed.

dbrattli (approved):
Thanks for fixing!

PR #674, opened 2022-12-11
Fix unit-tests

MainRo:
This is a complement to #668 to fix all unit-tests on python 3.10:

- Asyncio sleep does not accept the loop argument anymore.
- PEP479 changed the behavior of StopIteration raised by generators.

### DIFF
--- a/rx/internal/enumerable.py
+++ b/rx/internal/enumerable.py
@@ -24,11 +24,10 @@ class Enumerable(object):
 
             for value in self:
                 if n <= 0:
-                    raise StopIteration
+                    return
                 n -= 1
                 yield value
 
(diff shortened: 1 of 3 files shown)


MainRo:
@benhiller can you check if this is ok for you?

@dbrattli I pushed the tag 1.6.2 before running the tests (my bad). Can we remove it so that I push a new one after this is merged (I may have the rights to do it, not sure)?

benhiller:
Yep, these changes make sense to me!

dbrattli (approved):
Looks good!

dbrattli:
@MainRo Just push a new tag i.e 1.6.3

Issue #581, opened 2021-08-30
concat_map implementation or equivalent?


iyz91:
Coming from RxJS it was really convenient having the concatMap operator to have a pipeline of dynamic observables that wait for the previous to complete. I noticed that RxPY does not have an obvious equivalent (closest is flat_map but I can't have interleaving).

From what relevant docs/issues I could find, it seems that the following would achieve the same functionality but feels dirty. The `of(...)`'s are a stand-in for some code that would dynamically generate a new observable depending on the result from the previous.

```
of(1).pipe(
  map(lambda x: of(x + 1)),    # Dynamically generate observable
  merge(max_concurrent=1),     # Merge and wait to finish
  map(lambda x: of(x + 1)),    # Dynamically generate observable
  merge(max_concurrent=1)      # Merge and wait to finish
)
```

Is there a better way to achieve this currently?
Is there any plan to implement a concat_map operator/alias for this functionality?

iyz91:
Just as a reference for anyone needing the same, I implemented it myself as follows:
```python
# Standard Imports
from typing import Callable

# Third Party Imports
from rx.core import Observable
import rx.operators as ops
from rx.core.typing import Mapper


def concat_map(mapper: Mapper) -> Callable[[Observable], Observable]:
  def _concat_map(source: Observable) -> Observable:
    return source.pipe(
      ops.map(mapper),
      ops.merge(max_concurrent=1)
    )
  return _concat_map
```

Now you can use it easily, e.g.:
```python
source.pipe(
  concat_map(observable_factory),
  concat_map(another_factory),
  # ...
)
```

If anybody sees any issues with the above please kindly comment and let me know.

skewty:
This shows merged..  Is a reason this issue was left  open?

PR #678, opened 2022-12-26
Concat map

matiboy:
- tests inspired by [RxJS](https://github.com/ReactiveX/rxjs/blob/master/spec/operators/concatMap-spec.ts)
- implementation from #581 [iyz91](https://github.com/iyz91) 's code
- documentation inspired by [RxJS](https://rxjs.dev/api/operators/concatMap)


dbrattli (approved):
@matiboy Thanks for adding this. Looks great!

dbrattli:
One thing I suddenly realized is that we might not need the `_concat_map` file if we make the composition inside `__init__.py` e.g:

```py
return compose(operators. Map(project), operators. Merge(max_concurrent=1))
```

matiboy:
> One thing I suddenly realized is that we might not need the `_concat_map` file if we make the composition inside `__init__.py` e.g:
> 
> ```python
> return compose(map(project), merge(max_concurrent=1))
> ```

Sorry mis-read initially and kept the "verbose" definition. But yes, beautiful one-liner.

dbrattli:
Looks great. PS: another change we want to do is to borrow the `curry_flip` function from the Expression library to simplify all the operator implementations by removing all the nested functions. Take a look at https://github.com/cognitedata/Expression/blob/main/expression/collections/block.py#L602 Thus you can put `source` back as the first parameter, and curry-flip will move it as the last parameter curried (returning a function that takes it as the parameter). I can open an issue to discuss this more.

PR #682, opened 2022-12-31
[WIP] Refactor using curry_flip from Expression library

dbrattli:


matiboy:
@dbrattli If I'm going to make changes related to this like add curry_flip on some operators, maybe amending some types or adding overrides, would you prefer if I branch out and PR into this branch or push changes direct to the branch?

PR #683, opened 2022-12-31
Curry flip

matiboy:
Use curry flip on buffer_with_count
Propose some test cleanup (more readable)
Propose first draft for documentation on curry-flip usage
Propose (unnecessary?) typing overload for map_ when no mapper is provided
Hint on contributing - what checks to run

### DIFF
--- a/reactivex/operators/_buffer.py
+++ b/reactivex/operators/_buffer.py
@@ -2,13 +2,15 @@ from typing import Any, Callable, List, Optional, TypeVar
 
 from reactivex import Observable, compose
 from reactivex import operators as ops
+from reactivex.curry import curry_flip
+from reactivex.typing import UnaryOperator
 
 _T = TypeVar("_T")
 
 
 def buffer_(
     boundaries: Observable[Any],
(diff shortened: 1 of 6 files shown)


reactivex/operators/_buffer.py
```diff
@@ -2,13 +2,15 @@
...
+from reactivex.curry import curry_flip
+from reactivex.typing import UnaryOperator
 
 _T = TypeVar("_T")
 
 
 def buffer_(
     boundaries: Observable[Any],
-) -> Callable[[Observable[_T]], Observable[List[_T]]]:
+) -> UnaryOperator[_T, List[_T]]:
```
dbrattli: UnaryOperator?

dbrattli: Aha, now I see the definition below. Nice addition.

matiboy: Had to remove it because of a circular dependency. I'm not quite there with Python typing to be honest. Seems to confuse me, maybe cause I'm so used to TypeScript typing

dbrattli (approved):
Looks good!

PR #684, opened 2023-01-07
Switchmap operator

matiboy:
Minor docs amendments

Reviving the feature of switch_map proposed in #634 ; however I followed the [RxJS implementation](https://rxjs.dev/api/operators/switchMap) whereby:

1. The projector receives the map's index (use map_index instead of map)
2. There is no default projector. I don't feel that either `identity` or `of` is relevant, but would like to hear your thoughts on that since that was part of the conversation on #634  @giff-h

### DIFF
--- a/reactivex/operators/__init__.py
+++ b/reactivex/operators/__init__.py
@@ -3362,6 +3362,45 @@ def switch_latest() -> Callable[
     return switch_latest_()
 
 
+def switch_map(
+    project: Callable[[_T1, int], Observable[_T2]]
+) -> Callable[[Observable[_T1]], Observable[_T2]]:
+    """Projects each source value to an Observable which is merged in
+    the output Observable, emitting values only from the most recently
+    projected Observable.
+
+
+    .. marble::
+        :alt: switch_map
+
+        ---a----------b-------c---------------|
+        [  switch_map(x,i: x*i---x*i---x*i|)  ]
+        ---a---a---a--bb---bb-ccc---ccc---ccc-|
+
+    Examples:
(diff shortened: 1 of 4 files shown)


dbrattli:
1. Should we call it `switch_map_indexed`? This would align with `flap_map` vs `flat_map_indexed`
2. If you make the mapper optional, then you will automatically get the `identity` behaviour because of `map_idexed` where the mapper is optional.

reactivex/operators/__init__.py
```diff
@@ -3362,6 +3362,45 @@ def switch_latest() -> Callable[
     return switch_latest_()
 
 
+def switch_map(
+    project: Callable[[_T1, int], Observable[_T2]]
```
dbrattli, line 3366:
```suggestion
    project: Optional[MapperIndexed[_T1, Observable[_T2]]]
```

reactivex/operators/__init__.py
```diff
@@ -3362,6 +3362,45 @@ def switch_latest() -> Callable[
     return switch_latest_()
 
 
+def switch_map(
```
dbrattli, line 3365:
```suggestion
def switch_map_indexed(
```

matiboy: Sure, it's more consistent with RxPy itself. I'll set up both then I guess; a `switch_map` and a `switch_map_indexed`

matiboy: I'm not familiar with the `switch_starmap` operator proposed initially; if needed please let me know, I can look it up and set it up together with tests

dbrattli (approved):
Great work. Thanks for adding!

MainRo:
@dbrattli  can we merge this PR?

PR #686, opened 2023-02-02
Clarify what you can do with the .subscribe Disposable

tmewett:
Not certain this is right, please correct :)

`Disposable` nor `DisposableBase` are not documented so it wasn't clear what you could do with them.


MainRo:
Thanks, yes it good to precise it here. Especially because many small examples do not  show the disposal.

@dbrattli I do not understand why the code-quality check fails while they succeed on master.

matiboy:
@MainRo @dbrattli 

> @dbrattli I do not understand why the code-quality check fails while they succeed on master.

There appears to be an issue with isort/poetry which caused issues [on multiple repos](https://github.com/home-assistant/core/issues/86892) that depend on this pre-commit; it is fixed in 5.12 and in 5.11.5 (for python 3.7 compatibility)

https://github.com/PyCQA/isort/releases/tag/5.11.5

On #689 I attempted to bump isort to that version but my Poetry appears to be a more recent version and made a lot of changes so had to revert.

tmewett:
Can this be merged?

tmewett:
👀 ?

PR #689, opened 2023-02-07
Some operators revamped with curry flip

matiboy:
- take
- timestamp
- average

Additionally

1. Typed `run`'s return to T_out not Any
2. Removed `average`'s never hit check for count == 0 (take_last operator already raises) and fixed corresponding test
3. Try to clarify `all`'s behavior, specifically that it emits False and completes as soon as one item fails the predicate

### DIFF
--- a/reactivex/observable/observable.py
+++ b/reactivex/observable/observable.py
@@ -236,7 +236,7 @@ class Observable(abc.ObservableBase[_T_out]):
 
         return pipe_(self, *operators)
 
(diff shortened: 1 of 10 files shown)


dbrattli:
This is great 👍 One thing I just discovered is that PyCharm do not understand the `curry_flip` decorator. PyCharm is really bad with typing through decorators, so would we let that affect os or not?

dbrattli:
This will be fixed in #743

PR #690, opened 2023-02-16
Fix starmap_indexed not passing index

Boudewijn26:
This PR fixes `starmap_indexed` not actually including the index


reactivex/operators/_map.py
```diff
@@ -65,7 +65,7 @@ def _identity(value: _T1, _: int) -> _T2:
 
     return compose(
         ops.zip_with_iterable(infinite()),
-        ops.starmap_indexed(_mapper_indexed),
+        ops.starmap(_mapper_indexed),
```
Boudewijn26, line 68: This depended on the incorrect behavior of `starmap_indexed`. If it actually were to pass the index, it would pass it twice: once from `zip_with_iterable(infinite())` and once from `starmap_indexed`

tests/test_observable/test_starmap.py
```diff
@@ -405,6 +405,211 @@ def mapper(x, y):
         assert xs.subscriptions == [subscribe(200, 290)]
         assert invoked[0] == 3
 
+    def test_starmap_with_index_throws(self):
```
Boudewijn26, line 408: Tests derived from https://github.com/ReactiveX/RxPY/blob/8263b1e36231aeec956e825a991f3350ee3e74a7/tests/test_observable/test_map.py#L265 and adapted for starmap

PR #720, opened 2024-08-12
Introduced Observable >> operator as pipe shorthand

jackgene:
I was doing something with F# and ReactiveX, and noticed that because of F#'s `|>` operator, their pipelining syntax is a lot cleaner and more readable:
```fsharp
    let reqWithTimeout =
        req
        |> Observable.timeoutSpan (TimeSpan.FromSeconds 1)
        |> Observable.take 1
        |> Observable.catch Observable.empty
```

Here's the full [F# ReactiveX Application](https://github.com/jamesward/easyracer/blob/main/fsharp-reactive/EasyRacer/EasyRacer.fs).

So I thought that by overloading Python's right shift operator (`__rshift__`, `>>`), we could get something similar:
```python
        req() >> ops.timeout(datetime.timedelta(seconds=1)) >> ops.catch(rx.empty()),
```

Here's the [PR converting the entire RxPY application](https://github.com/jamesward/easyracer/pull/437/files) from `pipe`s to `>>`s.

While overloading the `>>` operator to perform pipelining seems like a strange choice, it is not without precedent, as [Airflow is doing the same thing](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/tasks.html) (note the `first_task >> second_task >> [third_task, fourth_task]` example).

WDYT?

### DIFF
--- a/reactivex/observable/observable.py
+++ b/reactivex/observable/observable.py
@@ -353,5 +353,21 @@ class Observable(abc.ObservableBase[_T_out]):
 
         return slice_(start, stop, step)(self)
 
+    def __rshift__(self, __op1: Callable[[Observable[_T_out]], _A]) -> _A:
+        """Shorthand for :func:`pipe <pipe>` (single argument overload).
+
+        Allows you to pipe through a single operator using `>>`.
+
+        Examples:
+            >>> source >> op == source.pipe(op)
+
+        Args:
+            operator: A single operator.
+
+        Returns:
+             The composed observable.
+        """
+        return self.pipe(__op1)
+
 
 __all__ = ["Observable"]


dbrattli:
This becomes too much "magic" imo. 
1. This is not the way things are normally done in Python
2. The operator `>>` means compose in F#, not pipe
3. Using `|` would perhaps be better but still too much "magic"

I would be more in favor of re-adding method-chaining syntax similar to C# 🙈 as an alternative to pipe. This is what I have done in the [Expression](https://github.com/dbrattli/expression) library which supports both pipe and method chaining.

jackgene:
> This becomes too much "magic" imo.
> 
> 1. This is not the way things are normally done in Python
> 2. The operator `>>` means compose in F#, not pipe
> 3. Using `|` would perhaps be better but still too much "magic"
> 
> I would be more in favor of re-adding method-chaining syntax similar to C# 🙈 as an alternative to pipe. This is what I have done in the [Expression](https://github.com/dbrattli/expression) library which supports both pipe and method chaining.

Re-introducing method chaining makes sense as well.

To be completely honest, the rationale for moving to pipes was never clear to me, and struck me as a step backwards. I thought it was just one of those things that made implementing RxPY easier.

MainRo:
Moving to pipes allowed implementing operators outside of the rxpy library. This is invaluable to me.
I always felt that operator overloading was awkward. This is used in many different languages but it always ends up being strange, and and non-standard related to these languages best-practices.

For several years, I have been thinking about writing a pep similar to the javascript proposal for a |> operator:
https://github.com/tc39/proposal-pipeline-operator

This js proposal has never been accepted, but most of the proposal applies to Python.
What do you think about going this way as the next big step ? I would be interested in co-writing such a pep if there is some interest.

jackgene:
Introducing a dedicated pipe-operator in Python seems like an interesting idea.

I'd be interested in chipping in, but would like to share some off the cuff thoughts. IMO, the crux of the problem isn't because function currying isn't Pythonic (and rarely used) like it is in ML-based language (such as F#), but rather a problem with `partial`ly applying functions.

Let me elaborate on what I mean. First of all, I think piping will work just fine with single-argument functions, so I don't think that'll go into that, other that a quick example (using `|>` since we all seem familiar with it):
```python
import math
import rand

math.sqrt(abs(random.randint(-1000, 1000)))

# Equivalent to:
random.randint(-1000, 1000) |> abs |> math.sqrt
```
Note that `abs`, and `math.sqrt` are both single-argument functions.

The problem becomes more interesting when dealing with function that takes more than one argument and needs to be partially applied:
```python
from functools import partial, reduc

reduce(
    lambda l, r: l + r,
    map(
        lambda n: n * 2,
        [1, 2, 3, 4, 5]
    )
)

# Equivalent to:
[1, 2, 3, 4, 5] \
|> partial(map, lambda n: n * 2) \
|> partial(reduce, lambda l, r: l + r)
```

Or worse, when you need to skip applied arguments (we need to apply the first and third argument to `reduce`, skipping the second):
```python
from functools import partial, reduc

reduce(
    lambda l, r: l + r,
    map(
        lambda n: n * 2,
        [1, 2, 3, 4, 5]
    ),
    0
)

# Equivalent to:
[1, 2, 3, 4, 5] \
|> partial(map, lambda n: n * 2) \
|> lambda nums: reduce(lambda l, r: l + r, nums, 0) # functools.partial does not work
```


So on the surface of it, I see this as requiring two PEPs:
1. Introduce the pipe-operator (called `|>`, or perhaps something more Pythonic, [Kotlin uses `let`, for instance](https://kotlin-quick-reference.com/158-R-let-operator.html). I feel like this would be the simpler proposal.
2. Introduce language-level support for partial function applications. I can think of two issues with `functools.partial` that needs addressing (though likely will be others):
      a. The syntax is somewhat verbose.
      b. It currently only partially applies the left-most parameters (so in my example above, you can partially apply the `reduce(function)` argument, but not `reduce(initial)`). You could argue that ML-based languages that rely on currying as its basis for partially applying functions have the same problem, but keep in mind that ML libraries tend to be designed with the pipe-operator in mind, and positions arguments with this in mind (hence F#'s [Sequence.fold](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-collections-seqmodule.html#fold) (the equivalent to `functools.reduce` in Python) has its `state` (aka `initial`) as its second argument). Python already has a rich ecosystem of libraries that may not necessarily have been designed with this in mind.

PR #722, opened 2024-11-07
Fix build by pining pyyaml to 5.3.1

MainRo:
see https://github.com/yaml/pyyaml/issues/724 for the origin of the error. We can remove this once pyyaml 7.0 is released.


MainRo:
@dbrattli the code-quality CI still fails. It passes locally on my setup. Can you have a look at it?

Kroppeb:
@MainRo It seems that there are also fixes to tests? Maybe those should be in a different PR?

dbrattli (approved):
LGTM

PR #736, opened 2025-10-15
refactor(combine-latest): simplify logic by removing redundant loops …

mrsants:
- Removed two unnecessary nested loops, reducing complexity and improving performance
- Replaced mutable list flags with `nonlocal` variables for cleaner state handling
- Added fallback `threading.RLock()` when `parent.lock` is not present
- Improved type hints using generics (`TypeVar`, `Tuple[T, ...]`)
- Added input validation to raise `ValueError` when no sources are provided
- Expanded and standardized docstring (PEP 257 + NumPy style)


mrsants:
@dbrattl how is the merge?

dbrattli (approved):

PR #742, opened 2025-11-08
docs: revise GIL section for Python 3.13 updates

timkofu:
Updated GIL explanation to include changes in Python 3.13 and free-threading mode.


dbrattli (approved):
Thanks for updating

PR #749, opened 2025-11-28
fix(gtk): call gi.require_version before importing gi.repository

selfisekai:
```
___________________________________________________ ERROR collecting tests/test_scheduler/test_mainloop/test_gtkscheduler.py ____________________________________________________
tests/test_scheduler/test_mainloop/test_gtkscheduler.py:16: in <module>
    gi.require_version("Gtk", "3.0")
/usr/lib/python3.12/site-packages/gi/__init__.py:121: in require_version
    raise ValueError(
E   ValueError: Namespace Gtk is already loaded with version 4.0
=============================================================================== warnings summary ================================================================================
tests/test_scheduler/test_mainloop/test_gtkscheduler.py:13
  /home/lauren/aports/community/py3-rx/src/RxPY-4.1.0/tests/test_scheduler/test_mainloop/test_gtkscheduler.py:13: PyGIWarning: Gtk was imported without specifying a version first. Use gi.require_version('Gtk', '4.0') before import to ensure that the right version gets loaded.
    from gi.repository import GLib, Gtk  # isort: skip
```


dbrattli:
Thanks for catching this — the fix is right, `require_version()` has to precede the first `from gi.repository import ...` or the namespace is already resolved by then.

I pushed two commits to your branch rather than asking you to rebase (hope that's alright — no force-push, your original commit is untouched):

- **Merge of `master`** — resolving one conflict in `tests/test_scheduler/test_mainloop/test_gtkscheduler.py`. `master` has since changed that import's comment to `# noqa: E402, I001  # type: ignore[import-untyped]`, so I kept your ordering with master's comment.
- **`fix(gtk): skip cleanly when GTK 3 is unavailable, pin Gdk in example`** — two follow-ups:
  - `require_version("Gtk", "3.0")` itself raises `ValueError` when GTK 3 isn't installed, so a GTK-4-only system still got a collection error, just one line further down. Now wrapped in `try`/`except ValueError` → `pytest.skip(..., allow_module_level=True)`. The tests call `Gtk.main()`/`Gtk.main_quit()`, removed in GTK 4, so skipping is the correct outcome there.
  - In the example, names in a single `from gi.repository import Gdk, GLib, Gtk` resolve left to right, so `Gdk` was loading before `Gtk` with no version pinned. Added `gi.require_version("Gdk", "3.0")`.

Also retitled the PR to satisfy the conventional-title check.

Worth noting for anyone reading later: PyGObject isn't in the test dependency group and no CI job installs GTK, so this file is skipped in CI entirely — this change is exercised only by downstream packagers like yourself. Which is to say the report was genuinely useful; we had no way to see it.

Will squash-merge once checks are green.

dbrattli (approved):

PR #793, opened 2026-05-11
docs: fix occured typo in run.py docstrings

SAY-5:
Fix "occured" -> "occurred" in two docstrings in reactivex/run.py.


dbrattli (approved):
Thanks, for fixing!

